PR #182, opened 2024-04-24
Implement a way for a holder of an Issuance to create a Presentation

### LINKED ISSUE #180: Add Use Case: Holder Presentation
I would like to use your library to create a **presentation as a holder**.

In the [KeyBindingTest](https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-sdjwt-kt/blob/main/src/test/kotlin/eu/europa/ec/eudi/sdjwt/KeyBindingTest.kt) there are code examples for the most relevant parts (see [class HolderActor](https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-sdjwt-kt/blob/66298d136a2057e4c17eea1964f4e639b0a023b8/src/test/kotlin/eu/europa/ec/eudi/sdjwt/KeyBindingTest.kt#L261)).

The only thing i'm missing is a sophisticating filtering of the disclosures that should be part of the SD-JWT presentation.
In your example, you are using a function `whatToDisclose: (Claim) -> Boolean` to filter the disclosures.
I see the problem that the objects neither of type `Claim` nor of type `Disclosure` have a context (path in the input claim set).
Without context, it is not possible to make a clear decision (same JSON key can appear several times).
 
I see two approaches to solving this problem:
- add context, e.g. JSON Path, to the disclosures -> `DisclosureWithPath` to enable proper filtering
- provide filter function that has a `releaseClaims` object, which is a subset of all claims, as input and outputs a list of all disclosures that belong to the `releaseClaims` (see example [OWF sd-jwt-kotlin library](https://github.com/openwallet-foundation-labs/sd-jwt-kotlin/blob/2f1aa2d09dae2a601b4ba15925333a95562f43dd/src/main/kotlin/org/sd_jwt/SdJwt.kt#L243))

Do you have any plans or more examples to implement proper filtering?

dzarras:
Laying the groundwork for implementing a query engine that given a SdJwt.Issuance and a Query, produces an appropriate SdJwt.Presentation.

TODO:

- [x] Remove the need for `parent` JsonPath. Given that the `SdClaimVisitor` is invoked from `RecreateClaims` we know that the JsonPath of the current node is `dot` notated. Thus the JsonPath of the parent can be easily calculated using the current JsonPath.
- [x] Update readme to include the new use case, including examples like PID or even more complex credentials from the specification 
- [x] Refactor the Key Binding Example (in tests) to use the high-level presentation function

Closes #180


babisRoutis:
@dzarras Please 
- Implement a value class named `JsonPointer` according to https://datatracker.ietf.org/doc/html/rfc6901
- Replace `SingleClaimJsonPath` class with `JsonPointer`

babisRoutis (approved):

PR #187, opened 2024-05-09
Replaced ECKey with AsymmetricJWK to get PublickKey

### LINKED ISSUE #186: KeyBindingJwt should be verified with RSA or EC keys
Currently KeyBindingJwt can only be verified with ECKey.

https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-sdjwt-kt/blob/6fd337bd38c70f4b42db9b6c70171f931efab71f/src/main/kotlin/eu/europa/ec/eudi/sdjwt/NimbusIntegration.kt#L123

If `holderPubKey` is RSA it will throw class-cast exception.

ydanneg:
Fixes #186

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/NimbusIntegration.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/NimbusIntegration.kt
@@ -21,6 +21,7 @@ import com.nimbusds.jose.JWSAlgorithm
 import com.nimbusds.jose.JWSHeader
 import com.nimbusds.jose.JWSSigner
 import com.nimbusds.jose.jwk.JWK
+import com.nimbusds.jose.jwk.AsymmetricJWK
 import com.nimbusds.jose.proc.SecurityContext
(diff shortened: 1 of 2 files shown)


babisRoutis (changes requested):
@ydanneg Perhaps the `holderPubKey` parameter should be declared of type `AsymmetricJWK`.
Can you please change the type (using an import alias of NimbusAsymmetricJWK) ?

babisRoutis:
@ydanneg Also please run `spotlessApply` to format code

babisRoutis:
@ydanneg Please note I am on Easter vacations. Will check this next week.

babisRoutis:
@ydanneg On second thought changing the parameter type will lead to some additional cascading changes.
I can push to your branch the changes if it is ok with you.

ydanneg:
> @ydanneg On second thought changing the parameter type will lead to some additional cascading changes.
> I can push to your branch the changes if it is ok with you.

Absolutely fine with it! :)

PR #190, opened 2024-05-13
Make verifications suspendable.

### LINKED ISSUE #189: Make verifications suspendable
Currently, top-level object `SdJwtVerifier` exposes various verification methods as plain non-suspendable functions.

This is fine as long as the caller has obtained beforehand (or out-of-band) the public key of the issuer.
There are cases though, like those described in SD-JWT-VC profile (see [here](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-03.html#name-issuer-signed-jwt-verificat)), where issuer's public key should be 
- Fetched (this is the case for an issuer that exposes an SD-JWT-VC [metadata](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-03.html#name-jwt-vc-issuer-metadata) endpoint) or/and
- Resolved (this is the case of an issuer that uses DIDs).
 
For this reason and in order to support in the future SD-JWT-VC   the methods of `SdJwtVerifier` must be suspendable.
This will also affect also the interfaces of `JwtSignatureVerifier` and `KeyBindingVerifier`

dzarras:
Closes #189

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
@@ -103,14 +103,14 @@ fun interface JwtSignatureVerifier {
      * @param jwt the JWT to validate
      * @return the payload of the JWT if signature is valid, otherwise raises [InvalidJwt]
      */
-    fun verify(jwt: String): Result<Claims> = runCatching { checkSignature(jwt) ?: throw InvalidJwt.asException() }
+    suspend fun verify(jwt: String): Result<Claims> = runCatching { checkSignature(jwt) ?: throw InvalidJwt.asException() }
 
     /**
      * Implement this method to check the signature of the JWT and extract its payload
      * @param jwt the JWT to validate
      * @return the payload of the JWT if signature is valid, otherwise null
      */
-    fun checkSignature(jwt: String): Claims?
+    suspend fun checkSignature(jwt: String): Claims?
 
     /**
      * Constructs a new [JwtSignatureVerifier] that in addition applies to the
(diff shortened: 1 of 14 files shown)


src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -119,7 +119,7 @@ fun interface JwtSignatureVerifier {
      * @return a new [JwtSignatureVerifier] that in addition applies to the
      *  extracted payload the [additionalCondition]
      */
-    fun and(additionalCondition: (Claims) -> Boolean): JwtSignatureVerifier = JwtSignatureVerifier { jwt ->
+    fun and(additionalCondition: suspend (Claims) -> Boolean): JwtSignatureVerifier = JwtSignatureVerifier { jwt ->
```
babisRoutis, line 122: Do we really need a suspended lamda here?

src/test/kotlin/eu/europa/ec/eudi/sdjwt/Sample.kt
```diff
@@ -66,7 +67,7 @@ fun genRSAKeyPair(): RSAKey =
         .issueTime(Date()) // issued-at timestamp (optional)
         .generate()
 
-fun main() {
+fun main() = runBlocking {
```
babisRoutis, line 70: You can just use `suspend fun main` instead

babisRoutis:
@dzarras  Perhaps we should rename milestone & target version to 0.4.0, since we are introducing a breaking change into the API

dzarras:
> @dzarras Perhaps we should rename milestone & target version to 0.4.0, since we are introducing a breaking change into the API

Done. Milestone renamed and version bumped.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
dzarras: Not necessarily, but I added it for API flexibility.

babisRoutis (approved):

PR #191, opened 2024-05-14
Add SD-JWT-VC verifier

### LINKED ISSUE #185: Verify issuance using trust anchor (CA certificate)
Hi.

Current version of the library including samples and tests verifies issuance with known Issuer's self-signed certificate only.
The real-world case will be that Issuer's certificate is signed by some trusted CA which is used as trust anchor by holder and verifiers.

CA (Trust Anchor) -> Issuer

[SD-JWT-VC](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-03.html#section-3.5) says it supports x5c JWT header that can be used to 
- verify chain (with known trust anchor)
- verify jwt signature

I think it should be applied to holders and verifiers that don't support `JWT VC Issuer Metadata` that can be used to retrieve Issuer public keys. Not sure about it. More specification analysis is required here.

It also can be that we don't know the type of a key used to sign issued credential JWT. 
Example can be that CA certificate is ECDSA and leaf certificate is RSA. Library should detect type algorithm from the header and use leaf certificate from x5c to verify a signature.

PS. I'm not expect in mentioned specification, probably some analysis is required additionally to support different cases.

Specs:
https://github.com/openid/oid4vc-haip-sd-jwt-vc/blob/main/openid4vc-high-assurance-interoperability-profile-sd-jwt-vc-1_0.md#issuer-identification-and-key-resolution-to-validate-an-issued-credential-issuer-key-resolution
https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-08.html#section-5.1

Code references:
https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-sdjwt-kt/blob/main/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt#L253
https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-sdjwt-kt/blob/main/src/main/kotlin/eu/europa/ec/eudi/sdjwt/NimbusIntegration.kt#L153

babisRoutis:
This PR adds a SD-JWT-VC specific variant of `SdJwtVerifier`

The implementation adds the ability to obtain Issuer's public key according to https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-03.html#name-issuer-signed-jwt-verificat

Library will be able to obtain issuer public 
- from SD-JWT-VC metadata
- from X5C header claim, provided that caller trusts the chain
- from DID, given that caller will provide a DID resolver (The DID resolution is out of scope of the lib)

Closes #185


src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
```diff
@@ -0,0 +1,229 @@
...
+}
+
+private suspend fun issuerJwsKeySelector(
+    trust: X509CertificateTrust?,
+    didResolver: DIDResolver?,
+    signedJwt: SignedJWT,
+    factory: KtorHttpClientFactory,
+): JWSKeySelector<SecurityContext>? {
+    val algorithm = requireNotNull(signedJwt.header.algorithm) { "missing 'alg'" }
+    return when (val source = keySource(signedJwt)) {
```
babisRoutis, line 110: @dzarras

Perhaps when() {}   could return a pub key and then assemble a `SingleKeyJWSKeySelector`

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/DnsUri.kt
```diff
@@ -0,0 +1,49 @@
...
+import io.ktor.http.*
+
+/**
+ * Gets the DNS name of the provided RFC-4501 DNS Uri.
+ *
+ * @return the DNS name or null in case [uri] is not a valid RFC-4501 DNS Uri
+ */
+internal fun dnsName(uri: Url): String? =
+    when (uri.protocol.name) {
+        "dns" -> {
```
babisRoutis, line 27: Could use a const

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
babisRoutis: Perhaps, `issuerJwsKeySelector` could be re-written as `issuerPubKey(source)`

dzarras: Can't do that. In case we have to fetch the Issuer's metadata, the metadata contain a JWKS/JWKS URI which can contain multiple public keys. Hence the need for constructing different JWSKeySelectors.

dzarras (dismissed):

dzarras:
@babisRoutis Unless you have other comments, we can go ahead and merge this.

babisRoutis:
@dzarras I was thinking that perhaps we should add a new section to the README (something like SD-JWT-VC support) pointing to the specification & just mentioning the SD-JWT-VC specific verifier & the KeyBindingTest example.

What do you think?

dzarras:
> @dzarras I was thinking that perhaps we should add a new section to the README (something like SD-JWT-VC support) pointing to the specification & just mentioning the SD-JWT-VC specific verifier & the KeyBindingTest example.
> 
> What do you think?

Sounds good. I'll update the README.

dzarras:
@babisRoutis Added a small section in README.

dzarras (approved):

PR #218, opened 2024-09-04
Update JWS JSON support

### LINKED ISSUE #217: Update support for an SD-JWT serialized in JWS JSON
Draft 9 introduced the following changes to JWS JSON serialization:

A new unprotected `header` was introduce which contains the `disclosures` and optionally the `kb_jwt`.  
This would be present both in the flattened and general options.

### LINKED ISSUE #219: Keep KB-JWT, if present, when verifying an SD-JWT presented to an RP
Currently, `SdJwtVerifier` provides the following method

```kotlin
suspend fun SdJwtVerifier.verifyPresentation(
    jwtSignatureVerifier: JwtSignatureVerifier,
    keyBindingVerifier: KeyBindingVerifier,
    unverifiedSdJwt: String,
): Result<SdJwt.Presentation<JwtAndClaims>>
```

The success outcome is a `SdJwt.Presentation` having the original JWT part of the SD-JWT and the presented disclosures.

There could be cases where caller would be interested to keep also the KB-JWT, if present, after the successful verification.

A possible method could be 

```kotlin
suspend fun verifyPresentationKeepKbJwt(
        jwtSignatureVerifier: JwtSignatureVerifier,
        keyBindingVerifier: KeyBindingVerifier,
        unverifiedSdJwt: String,
): Result<Pair<SdJwt.Presentation<JwtAndClaims>, JwtAndClaims?>> 
```

babisRoutis:
- [x] Update library to meet the new requirements for JWS JSON serialization
- [x] Add a new function to calculate KB-JWT and then serialize in JWS JSON for an `SdJwt.Presentation`
- [x] Add a new presentation verification function that expects the SD-JWT+KB in JWS JSON format.
- [x] Add similar methods to the SD-JWT-VC verifier
- [x] Add unit tests for all cases

Closes #217 
Closes #219

### DIFF
--- a/Module.md
+++ b/Module.md
@@ -3,7 +3,7 @@
 `eudi-lib-jvm-sdjwt-kt` offers a DSL (domain-specific language) for defining how a set of claims should be made selectively
 disclosable.
 
-Library implements [SD-JWT draft 8](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-08.html)
+Library implements [SD-JWT draft 12](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-12.html)
 is implemented in Kotlin, targeting JVM.
 
 Library's SD-JWT DSL leverages the DSL provided by
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ the [EUDI Wallet Reference Implementation project description](https://github.co
 This is a library offering a DSL (domain-specific language) for defining how a set of claims should be made selectively
 disclosable.
 
-Library implements [SD-JWT draft 8](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-08.html)
+Library implements [SD-JWT draft 12](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-12.html)
 is implemented in Kotlin, targeting JVM.
 
 Library's SD-JWT DSL leverages the DSL provided by
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ should be left outside the scope of this library.
 
 ## Presentation Verification
 
-### In simple (not enveloped) format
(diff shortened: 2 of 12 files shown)


dzarras:
@babisRoutis The code is OK. We can merge it as is. One comment for consideration in the future: I believe we are cluttering the API of the library with unneeded convenience methods.

For instance `SdJwtVerifier.verifyPresentationKeepingKbJwt` and/vs `SdJwtVerifier.verifyPresentation`. The later calls internally the first, and then maps the result of the first. Why not keeps only the later, with the signature of the first, and let the caller use destructing declarations to keep what he needs? i.e.
```kotlin
val (presentation, kbJwt) = SdJwtVerifier.verifyPresentation(...).getOrThrow()
val (presentation, _) = SdJwtVerifier.verifyPresentation(...).getOrThrow()
val (_, kbJwt) = SdJwtVerifier.verifyPresentation(...).getOrThrow()
```

dzarras (approved):

babisRoutis:
> @babisRoutis The code is OK. We can merge it as is. One comment for consideration in the future: I believe we are cluttering the API of the library with unneeded convenience methods.
> 
> For instance `SdJwtVerifier.verifyPresentationKeepingKbJwt` and/vs `SdJwtVerifier.verifyPresentation`. The later calls internally the first, and then maps the result of the first. Why not keeps only the later, with the signature of the first, and let the caller use destructing declarations to keep what he needs? i.e.
> 
> ```kotlin
> val (presentation, kbJwt) = SdJwtVerifier.verifyPresentation(...).getOrThrow()
> val (presentation, _) = SdJwtVerifier.verifyPresentation(...).getOrThrow()
> val (_, kbJwt) = SdJwtVerifier.verifyPresentation(...).getOrThrow()
> ```

Hi @dzarras 

I was thinking the same. 

Perhaps we can keep only the `verifyPresentation` and instead of 0.5.1, just go for 0.6.0

PR #220, opened 2024-09-17
fix: solve double tilda problem

VHerasymov:
### Double Tilde (~~) Appears When Serializing SD-JWT with Empty Disclosures

**Description**:
When calling the serialize function in the SdJwt class with an empty list of disclosures, the resulting serialized string ends with two tildes (~~). This appears to be an unintended consequence of how the concat function in StandardSerialization is designed.

**Steps to Reproduce**:
Create an SdJwt object with empty disclosures.
Call the serialize function.
Observe that the output ends with ~~.

**Cause of the Issue:**
The issue is caused by the concat function in the StandardSerialization object:

```
fun concat(
    serializedJwt: Jwt,
    disclosures: Iterable<String>,
): String {
    val serializedDisclosures = disclosures.concat { it }
    return "$serializedJwt$serializedDisclosures~"
}
```

When disclosures is empty, disclosures.concat { it } returns ~ due to the prefix in the concat function, leading to the double tilde (~~).

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtSerialization.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtSerialization.kt
@@ -77,7 +77,7 @@ internal object StandardSerialization {
         disclosures: Iterable<String>,
     ): String {
         val serializedDisclosures = disclosures.concat { it }
-        return "$serializedJwt$serializedDisclosures~"
+        return "$serializedJwt$serializedDisclosures"
     }
 
     /**
@@ -113,7 +113,7 @@ internal object StandardSerialization {
     }
 
     private fun <T> Iterable<T>.concat(get: (T) -> String): String =
-        joinToString(separator = "$TILDE", prefix = "$TILDE", transform = get)
+        joinToString(prefix = "$TILDE", separator = "") { "${get(it)}~" }
 }
 
 internal object JwsJsonSupport {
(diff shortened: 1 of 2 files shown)


babisRoutis:
Dear @VHerasymov
 
Thanks for this. 

Indeed that's a bug. 

Can I kindly ask you to add some simple unit tests  to your PR?
Otherwise, we can add those tests to your branch.

Please let me know and find bellow a minimal example


```kotlin
class SerializationTest {


    @Test
    fun sdJwtWithoutDisclosures() {
        val issuer = run {
            val issuerKey = ECKeyGenerator(Curve.P_256).generate()
            SdJwtIssuer.nimbus(signer = ECDSASigner(issuerKey), signAlgorithm = JWSAlgorithm.ES256)
        }
        val sdJwtSpec = sdJwt {
            plain {
                put("foo", "bar")
            }
        }
        val sdJwt = issuer.issue(sdJwtSpec).getOrThrow()
        val expected = "${sdJwt.jwt.serialize()}~"
        assertEquals(expected, sdJwt.serialize())
    }

    @Test
    fun sdJwtWithDisclosures() {
        val issuer = run {
            val issuerKey = ECKeyGenerator(Curve.P_256).generate()
            SdJwtIssuer.nimbus(signer = ECDSASigner(issuerKey), signAlgorithm = JWSAlgorithm.ES256)
        }
        val sdJwtSpec = sdJwt {
            plain {
                put("iss", "example")
            }
            sd {
                put("foo", "bar")
            }

        }
        val sdJwt = issuer.issue(sdJwtSpec).getOrThrow()
        val expected = buildString {
            append(sdJwt.jwt.serialize())
            append("~")
            for (d in sdJwt.disclosures) {
                append(d.value)
                append("~")
            }
        }
        assertEquals(expected, sdJwt.serialize())
    }

}
```

babisRoutis:
@VHerasymov 

Added to your branch the unit test

dzarras (approved):

dzarras:
@VHerasymov Thanks for spotting this 🥇

PR #224, opened 2024-10-07
Prefer java.util.Base64 over kotlin.io.encoding.Base64

### LINKED ISSUE #225: Make Base64-Url encoding/decoding consistent across Kotlin versions and strict
According to SD-JWT:
> Base64url denotes the URL-safe base64 encoding without padding defined in Section 2 of [[RFC7515](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-12.html#RFC7515)].

Currently the library uses Kotlin's `Base64.UrlSafe` to encode/decode values, and additionally exposes a method to remove padding characters. As of Kotlin 2.0.20 `Base64` default behavior changes, resulting in the library throwing exceptions when used from callers using Kotlin 2.0.20+. Reference: https://kotlinlang.org/docs/whatsnew2020.html#changes-to-the-base64-s-decoder-behavior

To avoid this discrepancy, the library must be updated to use an alternative well defined method to encode/decode Base64-URL, preferably Java's `Base64`.
Additionally the library must be updated to ensure that whenever Base64 Url-Encoded values are encountered, no padding is present, as per SD-JWT specification.

dzarras:
Replaces usages of Kotlin's Base64 with Java's Base64.

Closes #225

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/JwtBase64.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/JwtBase64.kt
@@ -15,17 +15,23 @@
  */
 package eu.europa.ec.eudi.sdjwt
 
-import kotlin.io.encoding.Base64
-import kotlin.io.encoding.ExperimentalEncodingApi
+import java.util.*
+import java.util.Base64.Decoder
+import java.util.Base64.Encoder
 
-@OptIn(ExperimentalEncodingApi::class)
 object JwtBase64 {
 
-    fun decode(value: String): ByteArray = Base64.UrlSafe.decode(value)
+    /**
+     * Base64 encoder that doesn't add padding.
+     */
+    private val encoder: Encoder = Base64.getUrlEncoder().withoutPadding()
 
-    // Since the complement character "=" is optional,
-    // we can remove it to save some bits in the HTTP header
-    fun encode(value: ByteArray): String = removePadding(Base64.UrlSafe.encode(value))
+    /**
+     * Base64 decoder that doesn't require padding.
+     */
+    private val decoder: Decoder = Base64.getUrlDecoder()
 
+    fun decode(value: String): ByteArray = decoder.decode(value)
+    fun encode(value: ByteArray): String = encoder.encodeToString(value)
     fun removePadding(value: String): String = value.replace("=", "")
 }
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/JwtBase64.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/JwtBase64.kt
@@ -21,17 +21,24 @@ import java.util.Base64.Encoder
 
 object JwtBase64 {
 
-    /**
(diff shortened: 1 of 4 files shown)


src/main/kotlin/eu/europa/ec/eudi/sdjwt/JwtBase64.kt
```diff
@@ -15,17 +15,23 @@
...
-    // we can remove it to save some bits in the HTTP header
-    fun encode(value: ByteArray): String = removePadding(Base64.UrlSafe.encode(value))
+    /**
+     * Base64 decoder that doesn't require padding.
+     */
+    private val decoder: Decoder = Base64.getUrlDecoder()
 
+    fun decode(value: String): ByteArray = decoder.decode(value)
+    fun encode(value: ByteArray): String = encoder.encodeToString(value)
     fun removePadding(value: String): String = value.replace("=", "")
```
babisRoutis, line 36: Can you think an alternative that doesn't expose the removePadding?

babisRoutis (changes requested):

src/main/kotlin/eu/europa/ec/eudi/sdjwt/JwtBase64.kt
dzarras: In the docs of `SdJwtDigest` we state that it represents a Base64 UrlSafe encoded value without padding. To me, it makes sense to move the responsibility of removing the padding to `SdJwtDigest` and keep `JwtBase64` only for encoding/decoding purposes.

ie.
```kotlin
package eu.europa.ec.eudi.sdjwt

import java.security.MessageDigest

/**
 * The digest of a [presentation][SdJwt.Presentation].
 * It contains the base64-url encoded digest of a presentation with all padding characters removed.
 */
@JvmInline
value class SdJwtDigest private constructor(val value: String) {

    companion object {
        const val CLAIM_NAME = "sd_hash"

        fun wrap(value: String): Result<SdJwtDigest> = runCatching {

//
//Remove padding below
//
            val clean = value.removePadding()
            JwtBase64.decode(clean)
            SdJwtDigest(clean)
        }

        fun digest(hashAlgorithm: HashAlgorithm, value: String): Result<SdJwtDigest> = runCatching {
            require(value.contains("~"))
            fun String.noKeyBinding() =
                if (endsWith("~")) {
                    this
                } else {
                    removeRange(lastIndexOf("~") + 1, length)
                }

            val digestAlgorithm = MessageDigest.getInstance(hashAlgorithm.alias.uppercase())
            val digest = digestAlgorithm.digest(value.noKeyBinding().encodeToByteArray())

//
//Remove padding below
//
            SdJwtDigest(JwtBase64.encode(digest).removePadding())
        }
    }
}

private fun String.removePadding() = replace("=", "")

```

babisRoutis: Can we find in the SD-JWT specification any requirements about padding for disclosures?

PR #230, opened 2024-10-16
Fix verification of SD-JWT VCs signed using EdDSA/OctetKeyPair

### LINKED ISSUE #229: Validating an EdDSA signed SD-JWT fails
I noticed that when using Ed25519 keys in their JWK form (`com.nimbusds.jose.jwk.OctetKeyPair`) the key selector does not find the key in question and the verification of the SD-JWT fails with an `InvalidJwt` error.

The source of the issue seems to be that nimbus `JWSVerificationKeySelector` does not work well with asymmetric keys that are not `ECKey` or `RSAKey` because when we get a match for a JWT that represents a OKP key (Ed25519 in our case), converting the key to Java keys via `com.nimbusds.jose.jwk.KeyConverter.toJavaKeys` just silently ignores the key in question.

I do realize that this is mostly in the nimbus library but as sdjwt-kt relies on this logic I wondered if you ever had a similar issue, or if you have some advice on how to approach solving it.

Thanks!

dzarras:
The main problem comes from Nimbus `DefaultJWTProcessor` which when processing a signed JWT :
- Requires all JWKs representing an issuer's pub keys to be representing a `java.security.Key`, yet
- `OctetKeyPair` doesn't implement the above conversion.

Solution provided by this PR, is to replace in the context of `sd-jwt-vc` verification the `DefaultJWTProcessor` with a custom processor, that 
- accepts a list of issuer pub keys (extracted by the sd-jwt-vc discovery process)
- instantiates an appropriate verifier (in the ED case there is Nimbus `Ed25519Verifier` that accepts the JWK)
- considers dig sig as valid, on the first verifier that succeeds.
- All other checks (for `typ`, claim set content etc, are preserved)

The above solution requires that caller includes on the class path Tink library, on its own.

Closes #229

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
@@ -15,25 +15,12 @@
  */
 package eu.europa.ec.eudi.sdjwt.vc
 
-import com.nimbusds.jose.JOSEObjectType
-import com.nimbusds.jose.JWSAlgorithm
-import com.nimbusds.jose.JWSVerifier
-import com.nimbusds.jose.crypto.ECDSAVerifier
-import com.nimbusds.jose.crypto.Ed25519Verifier
-import com.nimbusds.jose.crypto.MACVerifier
-import com.nimbusds.jose.crypto.RSASSAVerifier
 import com.nimbusds.jose.jwk.*
 import com.nimbusds.jose.jwk.source.ImmutableJWKSet
 import com.nimbusds.jose.jwk.source.JWKSource
-import com.nimbusds.jose.proc.BadJOSEException
-import com.nimbusds.jose.proc.DefaultJOSEObjectTypeVerifier
 import com.nimbusds.jose.proc.SecurityContext
 import com.nimbusds.jose.util.X509CertUtils
-import com.nimbusds.jwt.JWTClaimsSet
 import com.nimbusds.jwt.SignedJWT
-import com.nimbusds.jwt.proc.BadJWTException
-import com.nimbusds.jwt.proc.DefaultJWTClaimsVerifier
-import com.nimbusds.jwt.proc.DefaultJWTProcessor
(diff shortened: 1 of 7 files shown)


dzarras:
@babisRoutis Should we also promote Google's Tink to a runtime dependency?

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
```diff
@@ -272,14 +279,78 @@ internal fun keySource(jwt: SignedJWT): SdJwtVcIssuerPublicKeySource? {
...
 }
 
-private fun sdJwtVcProcessor(keySelector: JWSKeySelector<SecurityContext>): JWTProcessor<SecurityContext> =
-    DefaultJWTProcessor<SecurityContext>().apply {
+const val SD_JWT_VC_TYPE = "vc+sd-jwt"
+
+/**
+ * Custom [DefaultJWTProcessor] implementation with support for [OctetKeyPair].
+ */
+private class SdJwtVcJwtProcessor<C : SecurityContext>(
```
babisRoutis, line 287: Can you please extract the class in its own file, with `internal` scope

babisRoutis (approved):
@dzarras LGTM

Just a minor comment, to externalize on its file the custom processor and document (in the comments) why it was created.

babisRoutis:
> @babisRoutis Should we also promote Google's Tink to a runtime dependency?

I would say, no.

Instead we can document in the README, that the caller must include tink on its own, when dealing with ED.
I was thinking a simple knit example pointing to the test you created with the above disclaimer.

nemqe:
@babisRoutis I had a quick look LGTM

I could checkout this PR and try it out in our codebase if would find that valuable?

babisRoutis (approved):

babisRoutis:
> @babisRoutis I had a quick look and it LGTM.
> 
> I also checked-out the PR and tried it locally with the EdDSA signature - works like a charm.
> 
> Many thanks for the quick fix!

@nemqe  Thanks for the additional checks.

PR #235, opened 2024-11-06
Do not consider `kid` when selecting/filtering keys resolved from a DID Document

### LINKED ISSUE #234: JWK kid and DID interactions with Nimbus verifier
When verifying SD-JWTs that have been signed with a DID's key, Nimbus' `JWSVerificationKeySelector` - and therefore the SD-JWT verifier - needs the JWK found in the DID document to have a `kid` that coincides with the JWT header's `kid` to properly select the verification key.

With the current SD-JWT specifications, the JWT `kid` needs to be of the form `did:method:identifier#keyId` to allow for key resolution by holders and verifiers. ([section 3.5](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-sd-jwt-vc-05#section-3.5))

Verification can still be achieved in two ways: 
* by embedding such a `kid` in the JWK representation of verification methods in the DID document. But this is not standard: it is not mandatory to set that `kid` and the specification RECOMMENDS using the IETF format for `kid`s which conflicts with what the SD JWT VC spec requires (see [did-core spec](https://w3c.github.io/did-core/#verification-material)).
* by altering the JWK to set the appropriate `kid`, but any manipulation of the JWK is a bit unsavory.

To use the library's verifier in this setting, the DID has to be resolved and the correct key extracted beforehand. Is there a way the key selector can be used with a pre-selected key ? If not, should the library offer a verification entry point that does not rely on Nimbus selecting the key that has already been resolved to avoid manipulation of key material ?

dzarras:
Closes #234

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
@@ -15,8 +15,6 @@
  */
 package eu.europa.ec.eudi.sdjwt.vc
 
-import com.nimbusds.jose.JWSAlgorithm
-import com.nimbusds.jose.JWSHeader
 import com.nimbusds.jose.jwk.*
 import com.nimbusds.jose.jwk.source.ImmutableJWKSet
 import com.nimbusds.jose.jwk.source.JWKSource
@@ -211,7 +209,7 @@ private suspend fun issuerJwkSource(
     suspend fun fromDid(source: DIDUrl): JWKSource<SecurityContext>? =
         lookup
             ?.lookup(source.iss, source.kid)
-            ?.let { DIDJWKSet(signedJwt.header, JWKSet(it)) }
+            ?.let { SdJwtVcJwtProcessor.didJwkSet(signedJwt.header, JWKSet(it)) }
 
     return when (val source = keySource(signedJwt)) {
         null -> null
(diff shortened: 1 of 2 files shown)


src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
```diff
@@ -219,6 +221,40 @@ private suspend fun issuerJwkSource(
     }
 }
 
+/**
+ * [JWKSource] implementation for DID Documents.
+ *
+ * When [JWKSource.get] is invoked, it ignores the [JWKSelector] provided and instead one that matches all
+ * the properties of the provided [JWSHeader], but the Key ID.
+ */
+private class DIDJWKSet<C : SecurityContext>(jwsHeader: JWSHeader, val jwkSet: JWKSet) : JWKSource<C> {
```
babisRoutis, line 230: @dzarras Perhaps we should "nest" this class to the companion object of our processor, since this `JWKSource` can only be safely used from within this processor.

babisRoutis (approved):
@dzarras Minor comment

chumbert:
I tested this with our code and it works as expected.
My only change in our own code was disabling the `kid` override.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
dzarras: Could you check now? Is that OK?

babisRoutis (approved):

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
babisRoutis: Yes

PR #249, opened 2024-11-22
Allow a client to configure/enable its preferred method of SD-JWT-VC signing key resolution.

### LINKED ISSUE #248: Improve Issuer pub key discovery, while verifying an sd-jwt-vc
Currently, library expects that the issue of an sd-jwt-vc uses **one** of the three supported methods:
- SD-JWT-VC issuer metadata
- X.509 Certs 
- DID

Perhaps, the first two methods are not mutually exclusive.
This means that an issuer may choose to:
- Have as `iss` claim an HTTPS URI and optionally a `kid` claim in the JWS Header (enabling SD-JWT-VC Issuer metadata), and at the same time
- Have an `x5c` claim in the JWS header (enabling X509 Certs)

If that's the case, library should 
- stop treating like an error the above condition (where both methods are eligible)
- allow caller to indicate a preference in which method to use.

dzarras:
1. Makes the constructor of `SdJwtVcVerifier` private
~~2. Introduces a new fluent builder instantiated via `SdJwtVcVerifier.builder()` to allow clients to enable the desired resolution strategies. As a bare minimum the builder requires either SD-JWT-VC Issuer metadata or X509 Certificate trust to be configured. DID resolution is not required and can be enabled/configured alongside any other resolution strategy.~~
2. Introduces factory methods that allow the caller to enable the desired resolution strategies.

Closes #248

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,34 +395,67 @@ Example:
 <!--- INCLUDE
 import com.nimbusds.jose.JOSEObjectType
 import com.nimbusds.jose.JWSAlgorithm
-import com.nimbusds.jose.crypto.Ed25519Signer
+import com.nimbusds.jose.crypto.ECDSASigner
 import com.nimbusds.jose.jwk.Curve
-import com.nimbusds.jose.jwk.gen.OctetKeyPairGenerator
+import com.nimbusds.jose.jwk.gen.ECKeyGenerator
+import com.nimbusds.jose.util.Base64
+import com.nimbusds.jose.util.X509CertUtils
 import eu.europa.ec.eudi.sdjwt.*
 import eu.europa.ec.eudi.sdjwt.vc.SdJwtVcVerifier
 import kotlinx.coroutines.runBlocking
-import kotlin.io.encoding.Base64
-import kotlin.test.assertEquals
+import org.bouncycastle.asn1.DERSequence
+import org.bouncycastle.asn1.x509.Extension
+import org.bouncycastle.asn1.x509.GeneralName
+import org.bouncycastle.asn1.x509.GeneralNames
+import org.bouncycastle.cert.jcajce.JcaX509v3CertificateBuilder
+import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder
+import java.math.BigInteger
+import java.net.URL
+import java.time.Clock
+import java.util.*
+import javax.security.auth.x500.X500Principal
+import kotlin.time.Duration.Companion.days
+import kotlin.time.toJavaDuration
 -->
 
 ```kotlin
 val sdJwtVcVerification = runBlocking {
-    val key = OctetKeyPairGenerator(Curve.Ed25519).generate()
-    val didJwk = "did:jwk:${Base64.UrlSafe.encode(key.toPublicJWK().toJSONString().toByteArray())}"
+    val issuer = URL("https://issuer.example.com")
+    val key = ECKeyGenerator(Curve.P_521).generate()
+    val certificate = run {
+        val clock = Clock.systemDefaultZone()
+        val issuedAt = clock.instant()
+        val expiresAt = issuedAt + 365.days.toJavaDuration()
+        val subject = X500Principal("CN=${issuer.host}")
+        val signer = JcaContentSignerBuilder("SHA256withECDSA").build(key.toECPrivateKey())
+        val holder = JcaX509v3CertificateBuilder(
+            subject,
+            BigInteger.ONE,
+            Date.from(issuedAt),
+            Date.from(expiresAt),
+            subject,
+            key.toECPublicKey(),
+        ).addExtension(
+            Extension.subjectAlternativeName,
+            true,
+            GeneralNames.getInstance(DERSequence(GeneralName(GeneralName.dNSName, issuer.host))),
+        ).build(signer)
+        X509CertUtils.parse(holder.encoded)
+    }
 
     val sdJwt = run {
         val spec = sdJwt {
-            iss(didJwk)
+            iss(issuer.toExternalForm())
         }
-        val signer = SdJwtIssuer.nimbus(signer = Ed25519Signer(key), signAlgorithm = JWSAlgorithm.EdDSA) {
+        val signer = SdJwtIssuer.nimbus(signer = ECDSASigner(key), signAlgorithm = JWSAlgorithm.ES512) {
             type(JOSEObjectType("vc+sd-jwt"))
+            x509CertChain(listOf(Base64.encode(certificate.encoded)))
         }
         signer.issue(spec).getOrThrow()
     }
 
-    val verifier = SdJwtVcVerifier.usingDID { did, _ ->
-        assertEquals(didJwk, did)
-        listOf(key.toPublicJWK())
+    val verifier = SdJwtVcVerifier.usingX5c { chain ->
+        chain.isNotEmpty() && chain.first() == certificate
     }
 
     verifier.verifyIssuance(sdJwt.serialize())
--- a/README.md
+++ b/README.md
@@ -403,6 +403,7 @@ import eu.europa.ec.eudi.sdjwt.vc.SdJwtVcVerifier
 import kotlinx.coroutines.runBlocking
 import kotlin.io.encoding.Base64
 import kotlin.test.assertEquals
+import kotlin.test.fail
 -->
 
 ```kotlin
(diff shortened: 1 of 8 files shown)


dzarras:
@babisRoutis What do you think?

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
```diff
@@ -140,13 +140,42 @@ class SdJwtVcVerifier(
...
     }
 
-    private fun jwtSignatureVerifier(): JwtSignatureVerifier =
-        sdJwtVcSignatureVerifier(httpClientFactory, trust, lookup)
+    companion object {
+
+        /**
+         * Creates a new [SdJwtVcVerifier] with SD-JWT-VC Issuer Metadata resolution enabled, and optionally DID resolution enabled.
+         */
+        operator fun invoke(
```
babisRoutis, line 153: Given that Kotlin naturally supports default values to arguments, IMHO we should

- Either, have a single method `invoke`, accepting three optional arguments, all defaulting to `null`, or 
- Use (re-)name each `invoke` method. For instance, `usingX5cOrMetadata`, `usingX5c`, `usingMetadata`, `usingDID`  

In the later case, we can also mark the constructor as public (using `null` as default value for every attribute) and checking in the `init` that at least one is present.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
```diff
@@ -191,30 +220,32 @@ fun sdJwtVcSignatureVerifier(
...
 ): JWKSource<SecurityContext> {
     suspend fun fromMetadata(source: Metadata): JWKSource<SecurityContext> =
-        httpClientFactory.invoke().use { httpClient ->
+        httpClientFactory?.invoke()?.use { httpClient ->
             val fetcher = SdJwtVcIssuerMetaDataFetcher(httpClient)
             fetcher.fetchMetaData(source.iss)
                 ?.let { (_, jwks) -> ImmutableJWKSet(jwks) }
                 ?: invalidSdJwtVc("Failed to get SD-JWT-VC metadata of ${source.iss}")
-        }
+        } ?: invalidSdJwtVc("SD-JWT-VC Issuer metadata resolution is not enabled")
```
babisRoutis, line 234: Perhaps we could re-phrase it.  

> SD-JWT-VC validation requires Issuer's metadata resolution, which is not enabled

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
```diff
@@ -191,30 +220,32 @@ fun sdJwtVcSignatureVerifier(
...
-        if (trust.isTrusted(source.chain)) {
-            val jwk = JWK.parse(source.chain.first())
-            ImmutableJWKSet(JWKSet(mutableListOf(jwk)))
-        } else invalidSdJwtVc("Leaf certificate is not trusted")
+        trust?.let {
+            if (it.isTrusted(source.chain)) {
+                val jwk = JWK.parse(source.chain.first())
+                ImmutableJWKSet(JWKSet(mutableListOf(jwk)))
+            } else invalidSdJwtVc("Leaf certificate is not trusted")
+        } ?: invalidSdJwtVc("X509 Certificate trust is not enabled")
```
babisRoutis, line 242: Similarly

> SD-JWT-VC validation requires x5c validation, which is not enabled

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
```diff
@@ -191,30 +220,32 @@ fun sdJwtVcSignatureVerifier(
...
 
     suspend fun fromDid(source: DIDUrl): JWKSource<SecurityContext> =
-        lookup
-            ?.lookup(source.iss, source.kid)
-            ?.let { SdJwtVcJwtProcessor.didJwkSet(signedJwt.header, JWKSet(it)) }
-            ?: invalidSdJwtVc("Failed to resolve $source")
+        lookup?.let {
+            val jwks = it.lookup(source.iss, source.kid) ?: invalidSdJwtVc("Failed to resolve $source")
+            SdJwtVcJwtProcessor.didJwkSet(signedJwt.header, JWKSet(jwks))
+        } ?: invalidSdJwtVc("DID resolution is not enabled")
```
babisRoutis, line 248: > SD-JWT-VC validation requires DID resolution, which is not enabled

babisRoutis (changes requested):

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
```diff
@@ -140,13 +147,38 @@ class SdJwtVcVerifier(
...
+
+        /**
+         * Creates a new [SdJwtVcVerifier] with DID resolution enabled.
+         */
+        fun usingDID(didLookup: LookupPublicKeysFromDIDDocument): SdJwtVcVerifier = SdJwtVcVerifier(lookup = didLookup)
+
+        /**
+         * Creates a new [SdJwtVcVerifier] with SD-JWT-VC Issuer Metadata resolution enabled, X509 Certificate trust enabled.
+         */
+        fun usingIssuerMetadataOrX5c(
```
babisRoutis, line 177: Please rename to `usingX5cOrIssuerMetadata`, to indicate that `x5c` takes precedence

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
```diff
@@ -228,7 +262,7 @@ private suspend fun issuerJwkSource(
  */
 internal sealed interface SdJwtVcIssuerPublicKeySource {
 
-    data class Metadata(val iss: Url, val kid: String?) : SdJwtVcIssuerPublicKeySource
+    data class Metadata(val iss: Url) : SdJwtVcIssuerPublicKeySource
```
babisRoutis, line 265: Why was  `kid` removed?

Just double-checking this.

I think that we don't need it, since Nimbus would take it into consideration with its processor.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifierTest.kt
```diff
@@ -106,21 +107,14 @@ class SdJwtVcVerifierTest {
...
+        val expectedSource = SdJwtVcIssuerPublicKeySource.Metadata(Url("https://example.com"))
         testForMetaDataSource(expectedSource)
     }
 
     private fun testForMetaDataSource(expectedSource: SdJwtVcIssuerPublicKeySource.Metadata) {
         val jwt = run {
             val header = JWSHeader.Builder(JWSAlgorithm.ES256).apply {
                 type(JOSEObjectType(SD_JWT_VC_TYPE))
-                expectedSource.kid?.let { keyID(it) }
             }.build()
```
babisRoutis, line 118: Why was kid removed?

README.md
```diff
@@ -420,10 +420,11 @@ val sdJwtVcVerification = runBlocking {
         signer.issue(spec).getOrThrow()
     }
 
-    val verifier = SdJwtVcVerifier { did, _ ->
+    val verifier = SdJwtVcVerifier.usingDID { did, _ ->
```
babisRoutis, line 423: We need to change this example.
DID is not part of SD-JWT-VC

babisRoutis (changes requested):
Dimitris some more changes (minor).

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
dzarras: SD-JWT-VC doesn't say anything about `kid` during Issuer Metadata resolution.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifierTest.kt
dzarras: `kid` was populated using the value from `Metadata`. Since `Metadata` no longer contains `kid` it was removed. We could add it as an extra parameter to the method, if you believe it's required.

babisRoutis: Indeed there is no reference to `kid` in the resolution section.

There is though the following [recommendation](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-06.html#name-jwt-vc-issuer-metadata-resp):

>It is RECOMMENDED that the JWT contains a kid JWT header parameter that can be used to look up the public key in the JWK Set included by value or referenced in the JWT VC Issuer Metadata

I would say, to keep it, although in our current implementation we delegate this responsibility to nimbus.
An alternative, would be to peek from the JWK set returned via issuer's metadata, the corresponding JWK using the kid.

Please keep the `kid` and leave the alternative for another PR.

README.md
dzarras: Done.

babisRoutis (approved):

PR #253, opened 2024-11-27
Support SD-JWT-VC Type metadata

### LINKED ISSUE #240: Update media type to `application/dc+sd-jwt`
In draft SD-JWT-VC [draft 6](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-06.html) media type `application/vc+sd-jwt` was changed to `application/dc+sd-jwt`

Verifiers should accept SD-JWTs that use either type, in their header

### LINKED ISSUE #241: Support SD-JWT-VC Type Metadata
In draft 5 of SD-JWT-VC Type metadata were introduced. 

Library must support it,
- Model the type metadata
- Provide a way to resolve the metadata
- Incorporate the metadata to the verification of the an SD-JWT-VC (issued and/or presented)

### LINKED ISSUE #254: Deprecate the presentation functions that use `JsonPointer`
Currently, library used `JsonPointer` in order to 

- Allow a verifier to express "what to be disclosed" and 
- Allow a holder to using the disclosures to be included in the presented SD-JWT

With the introduction of SD-JWT-VC Type metadata and a `Claim Path` can be used instead of a Json Pointer.
The use of Claim Path has advantages:
- Can be used as a drop-in replaced while calculating the presentation. To put it simply, a Json Pointer has a one-to-one representation to `Claim Path`, with a single exception of the Json Poiner `""` (empty string, denoting the whole json element0). The resulting `Claim Path` will not have wildcard (`null`) elements.
- `Claim Path` is a more ergonomic way to express (as a verifier) what to be presented, since the support wildcard (`null`), a feature that is not supported by Json Pointer.

For these reasons:
- All `present()` methods that accept `JsonPointer`  will be marked as deprecated and removed in the future.
- A single present method, using `JsonPointer` will remain but named `presentJsonPointersMatching() ` or similar.
- New `present()` methods will be introduced that accept a `Set<ClaimPath>` that the presentation SD-JWT must satisfy.

babisRoutis:
This PR implements SD-JWT-VC Type metadata

- [x] Provide a model for Type metadata
- [x] Implement a matcher, that applies a `claim path` to a JSON Element
- [x] Use `ClaimPath` on `present()`
- [x] Deprecate present methods that use `JsonPointer`

Closes #240 
Closes #241 
Closes #254


src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtPresent.kt
```diff
@@ -89,24 +88,27 @@ fun <JWT> SdJwt.Issuance<JWT>.present(
...
+ * are relative to the unprotected JSON (not the JWT payload)
+ * @param claimsOf a function to obtain the [Claims] of the [SdJwt.jwt]
  * @receiver The issuance SD-JWT upon which the presentation will be based
+ * @param JWT the type representing the JWT part of the SD-JWT
  * @return the presentation if possible to satisfy the [query]
  */
-fun SdJwt.Issuance<JwtAndClaims>.present(
-    query: (JsonPointer) -> Boolean,
-): SdJwt.Presentation<JwtAndClaims>? = present(query) { (_, claims) -> claims }
+fun <JWT> SdJwt.Issuance<JWT>.present(
```
dzarras, line 99: @babisRoutis Would it make sense to keep two present methods? One where the query is a set of ClaimPaths (as is), and an analogous one where the query is a set of JsonPointers instead of a JsonPointer predicate?

src/main/kotlin/eu/europa/ec/eudi/sdjwt/Specs.kt
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Present only in an SD-JWT+KB, the Key Binding JWT
+     */
+    const val JWS_JSON_KB_JWT = "kb_jwt"
+
+    //
+    // Other
+    //
+    const val TILDE: Char = '~'
```
dzarras, line 59: @babisRoutis I would call this `DISCLOSURE_SEPARATOR` instead.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/Specs.kt
```diff
@@ -0,0 +1,189 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package eu.europa.ec.eudi.sdjwt
+
+/**
+ * [Selective Disclosure for JWTs (SD-JWT)](https://datatracker.ietf.org/doc/draft-ietf-oauth-selective-disclosure-jwt/)
+ */
+@Suppress("UNUSED")
+object SdJwtSpec {
```
dzarras, line 22: @babisRoutis Would it make sense to further segregate the constants inside nested objects and remove the prefixes? For instance `SdJwtSpec.CLAIM_SD` would become `SdJwtSpec.Claims.SD`.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/Specs.kt
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Digest of the SD-JWT to which the KB-JWT is tied
+     */
+    const val CLAIM_SD_HASH: String = "sd_hash"
+
+    /**
+     * Digest of the Disclosure for an array element
+     */
+    const val CLAIM_THREE_DOTS: String = "..."
```
dzarras, line 41: @babisRoutis Isn't this the claim name of an array element disclosure?

babisRoutis: Yes it is.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/Specs.kt
babisRoutis: No

src/main/kotlin/eu/europa/ec/eudi/sdjwt/Specs.kt
babisRoutis: Yes

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtPresent.kt
babisRoutis: Yes

PR #284, opened 2024-12-12
DSLv2

### LINKED ISSUE #283: DSL - redefinition
Currently DSL can be improved. Problems:
- Contains terms that didn't find their way to the specification like `recursive`, `structural` etc
- Contains terms that represent register IANA claims like `iss`, `aud` etc that don't fit well into such a DSL

Conceptually, the DSL needs to express a single property : Whether an element/attribute is always of selectively disclosable.

So a new version of the DSL could use two terms. 
- `sd` : To express a selectively disclosable value (former: `sd`) 
- `notSd` : The negation of it, which is always disclosable (former `plain`) 
For practical reasons two more pair of `terms` are defined:
- `sdObject` :  An selectively disclosable object that may contain always and/or selectively disclosable elements  (former: `recursice`)
- `notSdObject` :  Same as above but for an always disclosable object (former: `sturctured`) 
- `sdArray`: A selectively disclosable array that may contain always and/or selectively disclosable elements  (former: `recursiceArray`)
- `notSdArray`: Same as above, but for an always disclosable array (former: `sdArray` which was wrong)

The basic embellishment (`sd`, `notSd`) can be used to annotate:
- Any `JsonElement` (primitive, objects, arrays) 
- `DisclosableObject` a container of disclosable claims ( = JsonElement, DisclsoableObject, DisclosableArray)
- `DisclosableArray` a container of disclosable elements ( = JsonElement, DisclsoableObject, DisclosableArray) 

Some examples:

In the following example `address` will be represented with a singe disclosure. Either is being disclosed with all its attributes or not

```kotlin
    sdJwt {
        notSd("iss", "https://issuer.example.com")
        notSd("iat", 1683000000)
        notSd("exp", 1883000000)
        notSd("sub", "6c5c0a49-b589-431d-bae7-219122a9ec2c")

        sdObject("address") {
            notSd("street_address", "Schulstr. 12")
            notSd("locality", "Schulpforta")
            notSd("region", "Sachsen-Anhalt")
            notSd("country", "DE")
        }
    }
```

In the following example `address` is always disclosable, represented by 4 disclosures.  Each of its attributes can be selectively disclosed 

```kotlin
    sdJwt {
        notSd("iss", "https://issuer.example.com")
        notSd("iat", 1683000000)
        notSd("exp", 1883000000)

        notSdObject("address") {
            sd("street_address", "東京都港区芝公園４丁目２−８")
            sd("locality", "東京都")
            sd("region", "港区")
            sd("country", "JP")
        }
```

Finally, in the following example, `address` is selectively disclosable and the same is true of all its attributes. It will be represented by 5 disclosures.

```kotlin
    sdJwt {
        notSd("iss", "https://issuer.example.com")
        notSd("iat", 1683000000)
        notSd("exp", 1883000000)
        notSd("sub", "6c5c0a49-b589-431d-bae7-219122a9ec2c")

        sdObject("address") {
            sd("street_address", "Schulstr. 12")
            sd("locality", "Schulpforta")
            sd("region", "Sachsen-Anhalt")
            sd("country", "DE")
        }
    }
```

babisRoutis:
Closes #283


dzarras:
@babisRoutis Something else I noticed with the DSL is that we allow the redefinition of claims. i.e.:

```kotlin
        sdJwt {
            // plain claim in object
            notSd("plainClaim", "value")

            // sd claim in object
            sd("plainClaim", "value")
        }
```

In the example above the resulting spec will contain a single sd claim named `plainClaim`.

Should we keep this behavior, or would it be better to restrict it?

babisRoutis:
> @babisRoutis Something else I noticed with the DSL is that we allow the redefinition of claims. i.e.:
> 
> ```kotlin
>         sdJwt {
>             // plain claim in object
>             notSd("plainClaim", "value")
> 
>             // sd claim in object
>             sd("plainClaim", "value")
>         }
> ```
> 
> In the example above the resulting spec will contain a single sd claim named `plainClaim`.
> 
> Should we keep this behavior, or would it be better to restrict it?

@dzarras 

IMO, we shouldn't do anything. The sd-jwt-vc builder is actually a `Map` and that's also the case for kotlinx serialization builders.

dzarras (approved):

PR #288, opened 2024-12-14
Replace Java Base64 with kotlin.io

babisRoutis:
Towards a more kmp friendly code base, replaced Base64 implementation from Java with kotlin.io


dzarras (approved):

babisRoutis:
@dzarras Let's do not merge this right now. 

We need to make sure that caller is using Kotlin 2.x

dzarras:
Another alternative is [05nelsonm/encoding](https://github.com/05nelsonm/encoding), which provides stable multiplatform APIs for encoding.

dzarras (approved):

PR #291, opened 2024-12-15
Parameterized verifier

babisRoutis:
This PR parameterizes SdJwtVerifier to the representation of the JWT (of the SD-JWT)

```kotlin
interface SdJwtVerifier<out JWT> {

    suspend fun verifyIssuance(
        jwtSignatureVerifier: JwtSignatureVerifier<@UnsafeVariance JWT>,
        unverifiedSdJwt: String,
    ): Result<SdJwt.Issuance<JWT>>
    
    // other methods
    }
```
As a result, similarly, various more interfaces are also parameterized :
-  `JwtSignatureVerifier`
- `KeyBindingVerifier`
- `SdJwtVcVerifier`


src/main/kotlin/eu/europa/ec/eudi/sdjwt/NimbusIntegration.kt
```diff
@@ -70,7 +70,7 @@ import com.nimbusds.jwt.proc.JWTProcessor as NimbusJWTProcessor
  *
  * <em>Should not be used in production use cases</em>
  */
-internal val PlatformJwtSignatureVerifierNoSignatureValidation: JwtSignatureVerifier by lazy {
+internal val PlatformJwtSignatureVerifierNoSignatureValidation: JwtSignatureVerifier<Any> by lazy {
```
dzarras, line 75: Why `Any`? Isn't the type known to be `JsonObject` in this implementation?

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -260,7 +283,7 @@ typealias JwtAndClaims = Pair<Jwt, JsonObject>
  * A single point for verifying SD-JWTs in both [Combined Issuance Format][verifyIssuance]
  * and [Combined Presentation Format][verifyPresentation]
  */
-interface SdJwtVerifier {
+interface SdJwtVerifier<out JWT> {
```
dzarras, line 286: I'm starting to feel that the methods of `SdJwtVerifier` are not properly defined. 

Parameters like `jwtSignatureVerifier`, `keyBindingVerifier` feel more like an implementation detail rather than a runtime parameter.

By the way, if we have to mark `JWT` with `@UnsafeVariance` in so many instances, is it not the case that it is not an `out` parameter of `SdJwtVerifier`?

babisRoutis: Indeed, it seems that the `SdJwtVcVerifier` could be renamed to `SdJwtVerifierOps` and represent, in general, the operations to verify an sd-jwt (in string or JWS JSON). In this case, to obtain such an instance, caller would  have to provide a jwtSignatureVerifier and a kbJwtVerifier.

babisRoutis: > By the way, if we have to mark `JWT` with `@UnsafeVariance` in so many instances, is it not the case that it is not an `out` parameter of `SdJwtVerifier`?

Marked `SdJwtVerifier` as invariant to JWT

Issue #293, opened 2024-12-17
`SdJwtVcVerifier` must not always require KeyBinding JWT when verifying a presentation


dzarras:
`SdJwtVcVerifier` contains two methods that can be used to verify a presentation:

```kotlin
suspend fun verifyPresentation(
        unverifiedSdJwt: String,
        challenge: JsonObject? = null,
    ): Result<Pair<SdJwt.Presentation<JwtAndClaims>, JwtAndClaims?>>
```

and 

```kotlin
suspend fun verifyPresentation(
        unverifiedSdJwt: JsonObject,
        challenge: JsonObject? = null,
    ): Result<Pair<SdJwt.Presentation<JwtAndClaims>, JwtAndClaims?>>
```

Internally they both use `KeyBindingVerifier.forSdJwtVc(challenge)` to get an instance of `KeyBindingVerifier.MustBePresentAndValid` to validate the KeyBinding JWT.

A first issue that arises is the return type being incorrect. Given that the KeyBinding JWT must be present and valid, both of these methods must return `Result<Pair<SdJwt.Presentation<JwtAndClaims>, JwtAndClaims>>` with the KeyBinding JWT part being non-nullable.

Finally, and most importantly, the KeyBinding JWT is not required even during presentation. Quoting from the [spec](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-08.html#section-4.1):

> If the presentation of the SD-JWT VC includes a Key Binding JWT, the Key Binding JWT MUST adhere to the rules defined in Section 4.3 of [[I-D.ietf-oauth-selective-disclosure-jwt](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-08.html#I-D.ietf-oauth-selective-disclosure-jwt)].

Fixing the later will most probably introduce breaking API changes.

babisRoutis:
@dzarras  Good catch.

I suggest the following: 
- Correct the return type of the methods that you highlighted. Those would require the presence of KB-JWT 
- Consider introducing a new pair of methods that would enforce that no KB-JWT is present

babisRoutis:
> Fixing the later will most probably introduce breaking API changes.

We already moved main branch to 0.10.0-SNAPSHOT, that will introduce breaking changes

PR #296, opened 2024-12-17
Unify SD-JWT Issuance & Presentation into a single class

### LINKED ISSUE #295: Simplify SD-JWT representation
Currently, library represents SD-JWT as a sealed hierarchy comprised of two members `SdJwt.Issuance` and `SdJwt.Presentation`
Both of them having an identical content of two attributes: `jwt` and `disclosures`.

Given that library associates a KB-JWT, if needed, with a SD-JWT presented to an RP, during serialization/presentation the above distinction - between Issuance and Presentation - makes little sense.

Thus the following simplified representation can be used:

```kotlin
data class SdJwt<out JWT>(val jwt: JWT, val disclosures : List<Disclosure>)
```

Unfortunately, there is not an easy way to apply such a change, but to break pub API.

babisRoutis:
Closes #295


src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtPresent.kt
```diff
@@ -85,4 +85,4 @@ interface SdJwtPresentationOps<JWT> : SdJwtRecreateClaimsOps<JWT> {
 /**
  * Creates a Presentation that discloses **ALL** the claims of this Issuance.
  */
-fun <JWT> SdJwt.Issuance<JWT>.present(): SdJwt.Presentation<JWT> = SdJwt.Presentation(jwt, disclosures)
+fun <JWT> SdJwt<JWT>.present(): SdJwt<JWT> = SdJwt(jwt, disclosures)
```
dzarras, line 88: @babisRoutis This method can be safely removed.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -296,12 +288,12 @@ interface SdJwtVerifier<JWT> {
      * used to sign the SD-JWT
      * @param unverifiedSdJwt the SD-JWT to be verified
      * @return the verified SD-JWT, if valid. Otherwise, method could raise a [SdJwtVerificationException]
-     * The verified SD-JWT will contain a [JWT][SdJwt.Issuance.jwt] as both string and decoded payload
+     * The verified SD-JWT will contain a [JWT][SdJwt.jwt] as both string and decoded payload
      */
     suspend fun verifyIssuance(
```
dzarras, line 293: @babisRoutis I think this naming is no longer applicable. Maybe we should update the methods as follows:

* `suspend fun verify(jwtSignatureVerifier: JwtSignatureVerifier<JWT>, unverifiedSdJwt: String): Result<SdJwt<JWT>>` used to verify an SD-JWT
* `suspend fun verify(jwtSignatureVerifier: JwtSignatureVerifier<JWT>, keyBindingVerifier: KeyBindingVerifier<JWT>,, unverifiedSdJwt: String): Result<Pair<SdJwt<JWT>, JWT>>` used to verify an SD-JWT+KB

Similarly for methods that accept a JWS JsonObject.
Similarly for `SdJwtVcVerifier`.

This is also related to, and would also tackle, #293

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtPresent.kt
babisRoutis: @dzarras 
There could be an issue, if we remove this and it is related on how we have implemented the 
`fun present(query: Set<ClaimPath>): SdJwt<JWT>?` method.

If you pass an empty set to this method, then you get always a `null` result.

On the other hand the OpenId4VP DCQL defines the following semantics:
* Verifier may define a set of claim (paths), or even some alternative sets of claim (paths), to be presented.
* In case a Verifier doesn't provide such paths, then it is implied that the query is about all claims

So, to remove the method that you mention, IMHO, we need to change the behavior of the other method when passing an empty set.

What do you think?

babisRoutis: Removed the method.
Redefined the `present(query)` variant. In case of an empty query, returns the input SD-JWT

dzarras: @babisRoutis Agreed.

I propose the following:

`fun present(query: Set<ClaimPath>): Result<SdJwt<JWT>>`
* if query is empty, return SdJwt as is
* if query is non-empty:
  * if it contains irrelevant claim paths, fail
  * otherwise, create a new sd-jwt with the appropriate disclosures

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
babisRoutis: Agreed. Let's tackle this on another PR, addressing #293

dzarras (approved):

PR #302, opened 2024-12-19
Properly calculate the index of array elements during recreation.

### LINKED ISSUE #300: Incorrect `ClaimPath` for array elements when decoys are used
When recreating the claims of an SD-JWT, if the payload contains a `JsonArray` that contains Decoys, the `ClaimPaths` of the actual Array Elements are not properly calculated.
The current implementation always increments the array index, disregarding the fact that Decoys might have been employed in the array. [source](https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-sdjwt-kt/blob/main/src/main/kotlin/eu/europa/ec/eudi/sdjwt/RecreateClaims.kt#L225)

For instance the following with the following `JsonArray` as input:
```json
{
    "array": [
        "..decoy..",
        "..decoy..",
        "value",
        "..decoy.."
    ]
}
```

`value` is assigned the `ClaimPath` `['array', 2]`, when it should have been `['array', 0]`.

dzarras:
Fixes #300

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/RecreateClaims.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/RecreateClaims.kt
@@ -225,11 +225,15 @@ private class DiscloseObject(
     private fun discloseArray(
         currentPath: ClaimPath,
         jsonArray: JsonArray,
-    ): JsonArray {
+    ): JsonArray = buildJsonArray {
         var index = 0
-        return jsonArray
-            .mapNotNull { element -> discloseArrayElement(currentPath, element, index)?.also { index++ } }
-            .let { elements -> JsonArray(elements) }
+        jsonArray.forEach { element ->
+            discloseArrayElement(currentPath, element, index)
+                ?.let {
+                    add(it)
+                    index++
+                }
+        }
     }
 
     private fun discloseArrayElement(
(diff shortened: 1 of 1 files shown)


src/main/kotlin/eu/europa/ec/eudi/sdjwt/RecreateClaims.kt
```diff
@@ -225,11 +225,12 @@ private class DiscloseObject(
     private fun discloseArray(
         currentPath: ClaimPath,
         jsonArray: JsonArray,
-    ): JsonArray =
-        jsonArray
-            .zip(0..<jsonArray.size)
-            .mapNotNull { (element, index) -> discloseArrayElement(currentPath, element, index) }
+    ): JsonArray {
```
babisRoutis, line 228: @dzarras 
Personally I find somehow confusing the `?.also { }` which changes a state variable outside the the lambda.
Perhaps, a more imperative implementation gives a more clear view

```kotlin
buildJsonArray {
    var index = 0
    jsonArray.forEach { element ->
        discloseArrayElement(currentPath, element, index)
            ?.let { add(it); index++ }
    }
}
```

babisRoutis (dismissed):
@dzarras Approved. Just a minor comment for your consideration

PR #303, opened 2024-12-19
Makes various operations stack-safe

### LINKED ISSUE #299: Make operations stack-safe
The library contains some recursive operations that are currently not implemented in a stack-safe way.

Their implementation must be updated to make use of [DeepRecursiveFunction](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-deep-recursive-function/).

Targets:

- [x] [RecreateClaims](https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-sdjwt-kt/blob/main/src/main/kotlin/eu/europa/ec/eudi/sdjwt/RecreateClaims.kt)
- [x] [collectDigests](https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-sdjwt-kt/blob/main/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt#L527)
- [x] [SdJwtFactory](https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-sdjwt-kt/blob/main/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtFactory.kt)

dzarras:
Closes #299

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/RecreateClaims.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/RecreateClaims.kt
@@ -135,7 +135,7 @@ private class DiscloseObject(
     operator fun invoke(
         currentPath: ClaimPath?,
         jsonObject: JsonObject,
-    ): JsonObject = discloseObject(currentPath, jsonObject)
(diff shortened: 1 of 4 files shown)


babisRoutis (approved):
@dzarras Great !!

PR #337, opened 2025-05-14
Improve error message for InvalidDisclosures

### LINKED ISSUE #336: Improve error message for `InvalidDisclosures`
When trying to validate an SD-JWT/SD-JWT VC the disclosures of which incorrectly contain padding the `InvalidDisclosures` that is raised doesn't contain enough information about why the disclosure is invalid.

We should improve the error message. We could probably add a check to verify whether disclosures contain padding or not, and report a raise a new error.

An examples of an SD-JWT VC which disclosures that contain padding can be found here: https://github.com/eu-digital-identity-wallet/eudi-srv-web-verifier-endpoint-23220-4-kt/issues/349

atsant:
Closes #336

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 # Project properties
 group=eu.europa.ec.eudi
-version=0.13.1-SNAPSHOT
+version=0.14.0-SNAPSHOT
 
 # Kotlin configuration
 kotlin.code.style=official
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 # Project properties
 group=eu.europa.ec.eudi
-version=0.14.0-SNAPSHOT
+version=0.13.1-SNAPSHOT
 
 # Kotlin configuration
 kotlin.code.style=official
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
@@ -21,6 +21,7 @@ import eu.europa.ec.eudi.sdjwt.KeyBindingVerifier.MustBePresentAndValid
 import eu.europa.ec.eudi.sdjwt.KeyBindingVerifier.MustNotBePresent
 import eu.europa.ec.eudi.sdjwt.VerificationError.*
 import eu.europa.ec.eudi.sdjwt.vc.SdJwtVcVerificationError
+import io.ktor.util.*
 import kotlinx.serialization.json.*
 
 /**
@@ -52,7 +53,9 @@ sealed interface VerificationError {
      * SD-JWT contains invalid disclosures (cannot obtain a claim)
      */
     data class InvalidDisclosures(val invalidDisclosures: List<InvalidDisclosure>) : VerificationError { // List of object  opos afto sto 41
-        data class InvalidDisclosure(val disclose: String, val message: String? = null, val cause: Throwable? = null)
+        data class InvalidDisclosure(val disclosure: String, val message: String? = null, val cause: Throwable) {
+            // constructor(disclose: String, cause: Throwable) : this (disclose, null, cause)
+        }
     }
 
     /**
@@ -494,14 +497,22 @@ private fun verifyDisclosures(
  */
 private fun uniqueDisclosures(unverifiedDisclosures: List<String>): List<Disclosure> {
     val maybeDisclosures = unverifiedDisclosures.associateWith { Disclosure.wrap(it) }
-    maybeDisclosures.filterValues { it.isFailure }.keys.also { invalidDisclosures ->
+    maybeDisclosures.filterValues {
+        it.isFailure
+    }.also { invalidDisclosures ->
         if (invalidDisclosures.isNotEmpty()) {
-            val failedDisclosures: List<InvalidDisclosures.InvalidDisclosure> = invalidDisclosures.map {
-                InvalidDisclosures.InvalidDisclosure(it)
+            val error: List<InvalidDisclosures.InvalidDisclosure> = invalidDisclosures.map { (invalidDisclosure, failure) ->
+                val cause = failure.exceptionOrNull()!!
+                InvalidDisclosures.InvalidDisclosure(
+                    disclosure = invalidDisclosure,
+                    message = cause.message,
+                    cause = cause,
+                )
             }
-            throw InvalidDisclosures(failedDisclosures).asException()
+            throw InvalidDisclosures(error).asException()
         }
     }
+
     return unverifiedDisclosures.map { maybeDisclosures[it]!!.getOrThrow() }.also { disclosures ->
         if (maybeDisclosures.keys.size != disclosures.size) throw NonUniqueDisclosures.asException()
     }
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
@@ -51,7 +51,9 @@ sealed interface VerificationError {
     /**
      * SD-JWT contains invalid disclosures (cannot obtain a claim)
      */
-    data class InvalidDisclosures(val invalidDisclosures: List<String>) : VerificationError
+    data class InvalidDisclosures(val invalidDisclosures: List<InvalidDisclosure>) : VerificationError { // List of object  opos afto sto 41
+        data class InvalidDisclosure(val disclose: String, val message: String? = null, val cause: Throwable? = null)
+    }
 
     /**
      * SD-JWT contains a JWT which contains an unsupported Hashing Algorithm claim
@@ -493,8 +495,12 @@ private fun verifyDisclosures(
 private fun uniqueDisclosures(unverifiedDisclosures: List<String>): List<Disclosure> {
     val maybeDisclosures = unverifiedDisclosures.associateWith { Disclosure.wrap(it) }
     maybeDisclosures.filterValues { it.isFailure }.keys.also { invalidDisclosures ->
-        if (invalidDisclosures.isNotEmpty())
-            throw InvalidDisclosures(invalidDisclosures.toList()).asException()
+        if (invalidDisclosures.isNotEmpty()) {
+            val failedDisclosures: List<InvalidDisclosures.InvalidDisclosure> = invalidDisclosures.map {
+                InvalidDisclosures.InvalidDisclosure(it)
+            }
+            throw InvalidDisclosures(failedDisclosures).asException()
+        }
     }
     return unverifiedDisclosures.map { maybeDisclosures[it]!!.getOrThrow() }.also { disclosures ->
         if (maybeDisclosures.keys.size != disclosures.size) throw NonUniqueDisclosures.asException()
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
@@ -21,7 +21,6 @@ import eu.europa.ec.eudi.sdjwt.KeyBindingVerifier.MustBePresentAndValid
 import eu.europa.ec.eudi.sdjwt.KeyBindingVerifier.MustNotBePresent
 import eu.europa.ec.eudi.sdjwt.VerificationError.*
 import eu.europa.ec.eudi.sdjwt.vc.SdJwtVcVerificationError
-import io.ktor.util.*
 import kotlinx.serialization.json.*
 
 /**
@@ -52,10 +51,8 @@ sealed interface VerificationError {
     /**
      * SD-JWT contains invalid disclosures (cannot obtain a claim)
      */
-    data class InvalidDisclosures(val invalidDisclosures: List<InvalidDisclosure>) : VerificationError { // List of object  opos afto sto 41
-        data class InvalidDisclosure(val disclosure: String, val message: String? = null, val cause: Throwable) {
-            // constructor(disclose: String, cause: Throwable) : this (disclose, null, cause)
-        }
+    data class InvalidDisclosures(val invalidDisclosures: List<InvalidDisclosure>) : VerificationError {
+        data class InvalidDisclosure(val disclosure: String, val message: String? = null, val cause: Throwable)
     }
 
     /**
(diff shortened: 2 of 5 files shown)


src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -51,7 +52,11 @@ sealed interface VerificationError {
     /**
      * SD-JWT contains invalid disclosures (cannot obtain a claim)
      */
-    data class InvalidDisclosures(val invalidDisclosures: List<String>) : VerificationError
+    data class InvalidDisclosures(val invalidDisclosures: List<InvalidDisclosure>) : VerificationError { // List of object  opos afto sto 41
+        data class InvalidDisclosure(val disclosure: String, val message: String? = null, val cause: Throwable) {
+            // constructor(disclose: String, cause: Throwable) : this (disclose, null, cause)
```
dzarras, line 57: @atsant Please remove the commented out code.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -492,10 +497,22 @@ private fun verifyDisclosures(
...
 private fun uniqueDisclosures(unverifiedDisclosures: List<String>): List<Disclosure> {
     val maybeDisclosures = unverifiedDisclosures.associateWith { Disclosure.wrap(it) }
-    maybeDisclosures.filterValues { it.isFailure }.keys.also { invalidDisclosures ->
-        if (invalidDisclosures.isNotEmpty())
-            throw InvalidDisclosures(invalidDisclosures.toList()).asException()
+    maybeDisclosures.filterValues {
+        it.isFailure
+    }.also { invalidDisclosures ->
+        if (invalidDisclosures.isNotEmpty()) {
+            val error: List<InvalidDisclosures.InvalidDisclosure> = invalidDisclosures.map { (invalidDisclosure, failure) ->
```
dzarras, line 504: @atsant given this is a list, better name it `errors`. you can also remove the type annotation.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -21,6 +21,7 @@ import eu.europa.ec.eudi.sdjwt.KeyBindingVerifier.MustBePresentAndValid
 import eu.europa.ec.eudi.sdjwt.KeyBindingVerifier.MustNotBePresent
 import eu.europa.ec.eudi.sdjwt.VerificationError.*
 import eu.europa.ec.eudi.sdjwt.vc.SdJwtVcVerificationError
+import io.ktor.util.*
```
dzarras, line 24: @atsant why the ktor import?

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifierVerifyIssuanceTest.kt
```diff
@@ -118,6 +115,46 @@ class SdJwtVerifierVerifyIssuanceTest {
             QiLCAiY291bnRyeSI6ICJERSJ9XQ
     """.trimIndent().removeNewLine()
 
+    private suspend fun verifyIssuanceExceptingInvalidDisclosure(
+        list: List<String>,
```
dzarras, line 119: @atsant please use more descriptive names. i suspect this are the `invalidDisclosures`, correct?

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifierVerifyIssuanceTest.kt
```diff
@@ -118,6 +115,46 @@ class SdJwtVerifierVerifyIssuanceTest {
...
+                    val invalidDisclosures = assertIs<VerificationError.InvalidDisclosures>(exception.reason)
+                    assertEquals(list, invalidDisclosures.invalidDisclosures.map { it.disclosure })
+                } else {
+                    fail(exception.message)
+                }
+            },
+        )
+    }
+    private suspend fun verifyIssuanceExceptingInvalidDisclosure(
+        disclosures: List<String>,
```
dzarras, line 139: @atsant please rename to `invalidDisclosures`

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifierVerifyIssuanceTest.kt
```diff
@@ -118,6 +115,46 @@ class SdJwtVerifierVerifyIssuanceTest {
...
 
+    private suspend fun verifyIssuanceExceptingInvalidDisclosure(
+        list: List<String>,
+        unverifiedSdJwt: String,
+    ) {
+        val verification = DefaultSdJwtOps.verify(
+            jwtSignatureVerifier = DefaultSdJwtOps.NoSignatureValidation,
+            unverifiedSdJwt = unverifiedSdJwt,
+        )
+        verification.fold(
```
dzarras, line 126: @atsant You could reduce the code duplication. Try introducing a method `private fun Result<...>.validateInvalidDisclosures(invalidDisclosures: List<String>)` and call this method wherever needed.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifierVerifyPresentationTest.kt
```diff
@@ -218,6 +217,32 @@ class SdJwtVerifierVerifyPresentationTest {
...
+            val verification =
+                when (holderBindingVerifier) {
+                    KeyBindingVerifier.MustNotBePresent -> DefaultSdJwtOps.verify(jwtSignatureVerifier, unverifiedSdJwt)
+                    is KeyBindingVerifier.MustBePresentAndValid -> DefaultSdJwtOps.verify(
+                        jwtSignatureVerifier,
+                        holderBindingVerifier,
+                        unverifiedSdJwt,
+                    )
+                }
+            verification.getOrThrow()
```
dzarras, line 236: @atsant once you introduce the method i mentioned earlier, you can refactor this method to use it.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifierVerifyPresentationTest.kt
```diff
@@ -218,6 +217,32 @@ class SdJwtVerifierVerifyPresentationTest {
         }
     }
 
+    private suspend fun verifyPresentationExpectingDisclosureError(
+        disclosures: List<String>,
```
dzarras, line 221: @atsant i'm guessing `invalidDisclosures`

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifierVerifyIssuanceTest.kt
atsant: If we change this name to invalidDisclosures, then in line 150 it should be change to `invalidDisclosuresResults`

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
atsant: When importing a Throwable i missclicked on the ktor impl but I changed it immediately to the correct one.
Shouldn't Spotless catch unused imports?
Marked as resoloved.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifierVerifyPresentationTest.kt
```diff
@@ -218,6 +217,32 @@ class SdJwtVerifierVerifyPresentationTest {
...
+                        unverifiedSdJwt,
+                    )
+                }
+            verification.getOrThrow()
+            fail("Was expecting Disclosure error")
+        } catch (exception: SdJwtVerificationException) {
+            val invalidDisclosures = assertIs<VerificationError.InvalidDisclosures>(exception.reason)
+            assertEquals(disclosures, invalidDisclosures.invalidDisclosures.map { it.disclosure })
+
+//            assertEquals(expectedError, exception.reason)
```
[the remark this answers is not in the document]
atsant: I changed the assert and forgot the old one inside, my bad!

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
dzarras: I'm not really sure spotless checks for unused imports. It checks the formatting of the code. On the other hand, sonar detects unused imports if I recall correctly.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -51,7 +51,9 @@ sealed interface VerificationError {
     /**
      * SD-JWT contains invalid disclosures (cannot obtain a claim)
      */
-    data class InvalidDisclosures(val invalidDisclosures: List<String>) : VerificationError
+    data class InvalidDisclosures(val invalidDisclosures: List<InvalidDisclosure>) : VerificationError {
```
babisRoutis, line 54: @dzarras , @atsant 

There is also an alternative approach to this (simpler, IMHO).

Model `InvalidDisclosures`  as follows 

```kotlin
 data class InvalidDisclosures(val invalidDisclosures: Map<String, List<String>>) : VerificationError{
    init { 
        require(invalidDisclosures.size != 0) 
        require(invalidDisclosures.values.all { it.isNotEmpty() })
    }
    
    companion {
       operator fun invoke(reason: String, vararg invalidDisclosure: String): InvalidDisclosures = ....
    }
 } 
```
The key of the map is the reason to reject the disclosures (values of the map entry)
In this way, you can at the same time 
- Reject a single disclosure.  For instance, a disclosure is not properly encoded in base64
- Reject multiple disclosures for the same reason. For instance, not unique disclosures

atsant: The impl has been changed according to review.

While testing I found out that it is indeed working but there is a small issue/side-effect:

The message of the exception that we are using as the key contains information on where the error occurred.

ex: `Exception in thread "main" SdJwtVerificationException(reason=InvalidDisclosures(invalidDisclosures={The padding option is set to ABSENT, but the input has a pad character at index 62=[WyJ4WG8zQzRBYTFiN3ZEMWx6bTF3U2pnIiwiZ2l2ZW5fbmFtZSIsIkpvaG4iXQ==]}))
`

This means that for every different disclosure we will eventually get a new entry instead of being able to group them together.

We can use the name of the exception for this but I dont believe we actually provide more information with this way.

_Not recomended: A small hack can be to find if the word index exists and remove everything after it (??)_

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -492,10 +498,20 @@ private fun verifyDisclosures(
...
 private fun uniqueDisclosures(unverifiedDisclosures: List<String>): List<Disclosure> {
     val maybeDisclosures = unverifiedDisclosures.associateWith { Disclosure.wrap(it) }
-    maybeDisclosures.filterValues { it.isFailure }.keys.also { invalidDisclosures ->
-        if (invalidDisclosures.isNotEmpty())
-            throw InvalidDisclosures(invalidDisclosures.toList()).asException()
+    maybeDisclosures.filterValues {
+        it.isFailure
+    }.also { invalidDisclosures ->
+        if (invalidDisclosures.isNotEmpty()) {
+            val cause = invalidDisclosures.map { (invalidDisclosure, message) ->
```
dzarras, line 505: @atsant please rename the message variable. this is the failed Result. a more appropriate name would probably be `failure`.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
dzarras: Don't implement the workaround you mention. If the error message ever changes, it will break our implementation.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -492,10 +498,20 @@ private fun verifyDisclosures(
...
+                message.exceptionOrNull()!!.message!! to invalidDisclosure
```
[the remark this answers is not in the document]
atsant: I will not change it for now since https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-sdjwt-kt/pull/337#discussion_r2088812001 can solve both issues.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
babisRoutis: @atsant , @dzarras : Perhaps an enum to the InvalidDesclosures companion object (using the enum value as key of the map)  

Alternatively, 
- keep the string key, but expose error-specific factory methods (that would fix the key)
- Declare the keys (string) as public consts to the companion

Example
```kotlin
companion object {
    public const val INVALID_BASE64 = "...."
    fun invalidBase64(vararg disclose: String): InvalidDisclosures = ...
}
```

dzarras: OK let's try this out. We have to check the reason why a Disclosure could not be wrapped, and group together invalid Disclosures by failure reason. 

[Disclosure.wrap()](https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-sdjwt-kt/blob/main/src/main/kotlin/eu/europa/ec/eudi/sdjwt/Disclosure.kt#L67) might fail for a number of reasons, the value not being properly base64 url-safe encoded being one of them.

dzarras (dismissed):

gradle.properties
```diff
@@ -1,6 +1,6 @@
 # Project properties
 group=eu.europa.ec.eudi
-version=0.13.1-SNAPSHOT
+version=0.14.0-SNAPSHOT
```
dzarras, line 3: @atsant Please revert this change. The version must not be updated.

dzarras (approved):

PR #348, opened 2025-05-21
Add RetrieveTypeMetadata with default impl

### LINKED ISSUE #342: Implement retrieval of Type Metadata for VCTs that are an HTTPS URLs
Quoting the [spec](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-08.html#name-from-a-url-in-the-vct-claim):

> A URI in the `vct` claim can be used to express a type. If the type is a URL using the HTTPS scheme, Type Metadata MAY be retrieved from it.
The Type Metadata is retrieved using the HTTP GET method. The response MUST be a JSON object as defined in [Section 6.2](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-08.html#type-metadata-format).
If the claim `vct#integrity` is present in the SD-JWT VC, its value `vct#integrity` MUST be an "integrity metadata" string as defined in Section [Section 7](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-08.html#document-integrity).

- [x] Introduce a new functional interface `RetrieveTypeMetadata` as follows:
```kotlin
fun interface RetrieveTypeMetadata {
  operator suspend fun invoke(vct: Vct): Result<SdJwtVcTypeMetadata>
}
```

- [x] Introduce a new class `RetrieveTypeMetadataUsingKtor` that implements `RetrieveTypeMetadata` using Ktor, as described by the spec.

_**Integrity validation, recursive resolution of parents, and fetching of Json Schema by URI is currently out of scope.**_

atsant:
Relates to #342

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/RetrieveTypeMetadata.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/RetrieveTypeMetadata.kt
@@ -30,7 +30,7 @@ class RetrieveTypeMetadataUsingKtor(
     override suspend fun invoke(vct: Vct): Result<SdJwtVcTypeMetadata> = runCatching {
         val url = Url(vct.value)
         require(url.protocol == URLProtocol.HTTPS)
-        httpClientFactory().retrieveTypeMetadata(url)
+        httpClientFactory().use { httpClient -> httpClient.retrieveTypeMetadata(url) }
     }
 
     private suspend fun HttpClient.retrieveTypeMetadata(url: Url): SdJwtVcTypeMetadata =
@@ -41,7 +41,7 @@ class RetrieveTypeMetadataUsingKtor(
                         URLProtocol.HTTPS
                     }
                     headers {
-                        append("Accept", "application/json")
+                        append(HttpHeaders.Accept, ContentType.Application.Json)
                     }
                 }
             return httpResponse.body<SdJwtVcTypeMetadata>()
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/RetrieveTypeMetadata.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/RetrieveTypeMetadata.kt
@@ -38,6 +38,7 @@ class RetrieveTypeMetadataUsingKtor(
             url { URLProtocol.HTTPS }
             headers { append(HttpHeaders.Accept, ContentType.Application.Json) }
         }
+        check(httpResponse.status.isSuccess())
         return httpResponse.body<SdJwtVcTypeMetadata>()
     }
 }
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/RetrieveTypeMetadata.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/RetrieveTypeMetadata.kt
@@ -33,17 +33,11 @@ class RetrieveTypeMetadataUsingKtor(
         httpClientFactory().use { httpClient -> httpClient.retrieveTypeMetadata(url) }
     }
 
-    private suspend fun HttpClient.retrieveTypeMetadata(url: Url): SdJwtVcTypeMetadata =
-        use { httpClient ->
-            val httpResponse = httpClient
-                .get(url) {
-                    url {
-                        URLProtocol.HTTPS
-                    }
-                    headers {
-                        append(HttpHeaders.Accept, ContentType.Application.Json)
-                    }
-                }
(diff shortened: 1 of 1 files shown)


src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/RetrieveTypeMetadata.kt
```diff
@@ -23,18 +24,26 @@ fun interface RetrieveTypeMetadata {
...
         require(url.protocol == URLProtocol.HTTPS)
-        retrieveMetadata(url)
+        httpClientFactory().retrieveTypeMetadata(url)
     }
 
-    private suspend fun retrieveMetadata(url: Url): SdJwtVcTypeMetadata = httpClientFactory()
-        .use { httpClient ->
-            val httpResponse = httpClient.get(url)
+    private suspend fun HttpClient.retrieveTypeMetadata(url: Url): SdJwtVcTypeMetadata =
+        use { httpClient ->
```
dzarras, line 37: @atsant do not invoke `use` on extensions functions. invoke `use` in the place you are creating the `HttpClient`. If you want to use the same `HttpClient` multiple times, if you invoke `retrieveTypeMetadata` as is, it will close the `HttpClient` and you will not be able to reuse it later on. 

As a rule of thumb: closing a resource is the responsibility of the one who created it.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/RetrieveTypeMetadata.kt
```diff
@@ -23,18 +24,26 @@ fun interface RetrieveTypeMetadata {
...
+            val httpResponse = httpClient
+                .get(url) {
+                    url {
+                        URLProtocol.HTTPS
+                    }
+                    headers {
+                        append("Accept", "application/json")
```
dzarras, line 44: @atsant please use the constants in `HttpHeader` and `MediaType` (if i'm not mistaken) provided by Ktor.

atsant: Yes already on it 🫡

dzarras: https://api.ktor.io/ktor-http/io.ktor.http/-http-headers/index.html

dzarras: https://api.ktor.io/ktor-http/io.ktor.http/-content-type/index.html

dzarras (changes requested):

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/RetrieveTypeMetadata.kt
atsant: So the extension function call should be 
```kotlin 
httpClientFactory().use { httpClient -> httpClient.retrieveTypeMetadata(url) })
```
right?

dzarras: exactly

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/RetrieveTypeMetadata.kt
```diff
@@ -0,0 +1,43 @@
...
+        require(url.protocol == URLProtocol.HTTPS)
+        httpClientFactory().use { httpClient -> httpClient.retrieveTypeMetadata(url) }
+    }
+
+    private suspend fun HttpClient.retrieveTypeMetadata(url: Url): SdJwtVcTypeMetadata {
+        val httpResponse = get(url) {
+            url { URLProtocol.HTTPS }
+            headers { append(HttpHeaders.Accept, ContentType.Application.Json) }
+        }
+        return httpResponse.body<SdJwtVcTypeMetadata>()
```
dzarras, line 41: @atsant one last minor comment. before getting the body of the response, do a `check` that the status of the response is 200/ok.

atsant: If the request failes, http client throws an ClientRequestException with the http code inside (ex 400 Bad request) and it doesnt proceed.

In case of the wrong payload it returns JsonConvertException.

If we need to actually check the response 200 Kotlin asks for an extra when default clause.

Same with if ( needs an else branch).

In both cases how should this be handled ?

dzarras (approved):

PR #362, opened 2025-06-02
DSL Code re-arrangment 

babisRoutis:
- Yet another re-arrangment of pacakges and classes
- Removed SdJwtDefinition.claimPaths()  method. It was not used 
- Cleanup tests, removed obsolete code
- Added a seperate hierarchy for "definitions" distinct from "values". Kept the Alt (OneOf or AnyOne) 
- Re-implementing the validator. Gone is the DeepRecusrsiveScope method.
- Added a new method named SdJwtDefinition.fromSdJwtVcMetadataStrict() that "parses" the type metadata with some assumptions to mitigate the under-defined arrays of the spec. Kept the the old function as well


src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/DisclosableElementDefinition.kt
```diff
@@ -0,0 +1,120 @@
...
+     * the element's value must match at least one of the provided [DisclosableElementDefinition]s.
+     *
+     * @param value A [Set] of [DisclosableElementDefinition]s, where the data element must satisfy at least one.
+     */
+    @JvmInline
+    value class Alt<K, out A> (
+        val value: Set<DisclosableElementDefinition<K, A>>,
+    ) : DisclosableDef<K, A> {
+        init {
+            require(value.size >= 2) { "At least 2 values must be provided" }
```
dzarras, line 117: @babisRoutis Maybe we should also limit the content of Alt. I was thinking about disallowing the creation of an Alt that contains another Alt.

Issue #347, opened 2025-05-20
SD-JWT VC verifier to validate recreated/process payload per Json Schema in Type metadata


dzarras:
Given an SD-JWT VC that has passed the initial set of validations, and its Type metadata:

If the Type metadata contain a Json Schema, then SD-JWT VC Verifier must validate the recreated/process payload of the SD-JWT VC against the Json Schema in the Type metadata.

PR #366, opened 2025-06-04
Add support for Json Schema validation in `SdJwtVcVerifier`

dzarras:
1. Introduces `JsonSchemaValidator`. A functional interface that exposes a SAM to validate a Json payload against a Json Schema.
2. Add an optional `JsonSchemaValidator` parameter to `SdJwtVcVerifierFactory`
3. Add support for Json Schema validation in `NimbusSdJwtVcVerifier` when a `JsonSchemaValidator` has been provided
4. Introduces new error codes in `SdJwtVcVerificationError` that relate to Json Schema validation

Relates to #347

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/DefinitionBasedSdJwtVcValidator.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/DefinitionBasedSdJwtVcValidator.kt
@@ -79,9 +79,27 @@ sealed interface DefinitionViolation {
     data class IncorrectlyDisclosedClaim(val claimPath: ClaimPath) : DefinitionViolation
 }
 
+/**
+ * The validation result of an SD-JWT VC against an SdJwtDefinition.
+ */
 sealed interface DefinitionBasedValidationResult {
-    data class Valid(val payload: JsonObject) : DefinitionBasedValidationResult
 
+    /**
+     * The SD-JWT VC was successfully validated.
+     *
+     * @property recreatedCredential the processed/recreated payload of the successfully validated SD-JWT VC
+     * @property disclosuresPerClaimPath disclosures per claim path
+     */
+    data class Valid(
+        val recreatedCredential: JsonObject,
+        val disclosuresPerClaimPath: DisclosuresPerClaimPath,
+    ) : DefinitionBasedValidationResult
+
+    /**
+     * The SD-JWT VC could not be successfully validated.
+     *
+     * @property errors the definition violations that were detected
+     */
     data class Invalid(val errors: List<DefinitionViolation>) : DefinitionBasedValidationResult {
         constructor(
             head: DefinitionViolation,
@@ -299,7 +317,7 @@ private class SdJwtVcDefinitionValidator private constructor(
             sdJwt: UnsignedSdJwt,
             definition: DisclosableDefObject<String, *>,
         ): DefinitionBasedValidationResult {
-            val (processedPayload, disclosuresPerClaimPath) = runCatching {
+            val (recreatedCredential, disclosuresPerClaimPath) = runCatching {
                 val disclosuresPerClaimPath = mutableMapOf<ClaimPath, List<Disclosure>>()
                 val visitor = disclosuresPerClaimVisitor(disclosuresPerClaimPath)
                 sdJwt.recreateClaims(visitor) to disclosuresPerClaimPath.toMap()
@@ -307,9 +325,9 @@ private class SdJwtVcDefinitionValidator private constructor(
                 return DefinitionBasedValidationResult.Invalid(DefinitionViolation.DisclosureInconsistencies(it))
             }
 
-            val errors = SdJwtVcDefinitionValidator(disclosuresPerClaimPath, definition).validate(processedPayload)
+            val errors = SdJwtVcDefinitionValidator(disclosuresPerClaimPath, definition).validate(recreatedCredential)
             return if (errors.isEmpty()) {
-                DefinitionBasedValidationResult.Valid(processedPayload)
+                DefinitionBasedValidationResult.Valid(recreatedCredential, disclosuresPerClaimPath)
             } else {
                 DefinitionBasedValidationResult.Invalid(errors)
             }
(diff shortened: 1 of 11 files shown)


src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/NimbusSdJwtVcVerifierFactory.kt
```diff
@@ -86,18 +87,39 @@ private class NimbusSdJwtVcVerifier(
...
     private suspend fun ResolveTypeMetadata.resolve(sdJwt: SdJwt<NimbusSignedJWT>): ResolvedTypeMetadata =
         try {
             val vct = Vct(sdJwt.jwt.jwtClaimsSet.getStringClaim(SdJwtVcSpec.VCT))
             this(vct).getOrThrow()
         } catch (error: Exception) {
             raise(SdJwtVcVerificationError.TypeMetadataVerificationError.TypeMetadataResolutionFailure(error))
         }
+
+    private suspend fun JsonSchemaValidator.validate(sdJwt: SdJwt<NimbusSignedJWT>, schemas: List<JsonSchema>) {
+        val recreated = with(NimbusSdJwtOps) { sdJwt.recreateClaims(null) }
```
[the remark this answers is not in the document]
babisRoutis: I think that there could be two approaches to this

A) Include to the responsibilities of `DefinitionBasedSdJwtVcValidator` to check compliance of the recreated credential to a list of JSON schemas . In practice this means that when obtaining an instance of `DefinitionBasedSdJwtVcValidator` usually by `ResolvedTypeMetadata` we can provide also a way to validate against JSON Schema and use it accordingly.

B) DefinitionBasedSdJwtVcValidator evaluates a binary result representing by `DefinitionBasedValidationResult`. We can change `Valid` representation which currently is a `data object` to contain a "proof" that the credential was indeed valid in the form of the recreated credential. That is, the recreated credential is a bi-product of the definition-based validation. In this case, we can move the responsibility of checking against the list of schemas to the SdJwtVcVerifier by allowing the user to define yet another capability

src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/DefinitionBasedSdJwtVcValidator.kt
```diff
@@ -80,7 +80,7 @@ sealed interface DefinitionViolation {
 }
 
 sealed interface DefinitionBasedValidationResult {
-    data object Valid : DefinitionBasedValidationResult
+    data class Valid(val payload: JsonObject) : DefinitionBasedValidationResult
```
babisRoutis, line 83: - A better name, maybe?
- At least a KDodc
- What about including also the disclosure map?

src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/DefinitionBasedSdJwtVcValidator.kt
```diff
@@ -324,7 +309,7 @@ private class SdJwtVcDefinitionValidator private constructor(
 
             val errors = SdJwtVcDefinitionValidator(disclosuresPerClaimPath, definition).validate(processedPayload)
             return if (errors.isEmpty()) {
-                DefinitionBasedValidationResult.Valid
+                DefinitionBasedValidationResult.Valid(processedPayload)
```
babisRoutis, line 312: Use consistent names please. Perhaps, recreatedCredential or similar

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/NimbusSdJwtVcVerifierFactory.kt
```diff
@@ -75,8 +77,7 @@ private class NimbusSdJwtVcVerifier(
     override suspend fun verify(unverifiedSdJwt: String): Result<SdJwt<NimbusSignedJWT>> =
         NimbusSdJwtOps.verify(jwtSignatureVerifier, unverifiedSdJwt)
             .alsoCatching {
-                val typeMetadata = resolveTypeMetadata?.resolve(it)
-                typeMetadata?.validate(it)
+                validate(it)
```
babisRoutis, line 80: Can you please convert it to a block
and make two concrete calls 

It is way more readable than using also {} or similar

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/JsonSchemaValidator.kt
```diff
@@ -0,0 +1,89 @@
...
+     * Validation succeeded
+     */
+    data object Valid : JsonSchemaValidationResult
+
+    /**
+     * Validation failed.
+     *
+     * @property errors the violations that were detected
+     */
+    data class Invalid(val errors: List<JsonSchemaViolation>) : JsonSchemaValidationResult {
```
babisRoutis, line 49: Consider representing errors as Map<Int, List<JsonSchemaViolation>>
Where the key is the schema no.

Alternatively, just add the schemaNo to the violation

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/NimbusSdJwtVcVerifierFactory.kt
```diff
@@ -86,18 +87,40 @@ private class NimbusSdJwtVcVerifier(
...
             val vct = Vct(sdJwt.jwt.jwtClaimsSet.getStringClaim(SdJwtVcSpec.VCT))
             this(vct).getOrThrow()
         } catch (error: Exception) {
             raise(SdJwtVcVerificationError.TypeMetadataVerificationError.TypeMetadataResolutionFailure(error))
         }
+
+    private suspend fun JsonSchemaValidator.validatePayloadAgainst(payload: JsonObject, schemas: List<JsonSchema>) {
+        val result = try {
+            validate(payload, schemas)
+        } catch (error: Exception) {
```
babisRoutis, line 116: Why this try catch?

babisRoutis (approved):

PR #367, opened 2025-06-06
Added some helper function for SdJwtDefinition

babisRoutis:
The PR adds two convenient methods when working with SD-JWT-VC definitions

**Find element**

```kotlin
fun DisclosableDefObject<String, AttributeMetadata>.findElement(
    claimPath: ClaimPath,
): SdJwtElementDefinition?
```
Function requires that the given `claimPath` doesn't contain wildcard or array index elements

**Attribute metadata of a defintion**

```kotlin
fun DisclosableDef<String, AttributeMetadata>.attributeMetadata(): AttributeMetadata
```

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/SdJwtDefinition.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/SdJwtDefinition.kt
@@ -124,7 +124,6 @@ fun DisclosableDef<String, AttributeMetadata>.attributeMetadata(): AttributeMeta
         is DisclosableDef.Arr<String, AttributeMetadata> -> {
             check(value is SdJwtArrayDefinition)
             value.metadata
-            value.metadata
         }
 
         is DisclosableDef.Obj<String, AttributeMetadata> -> {
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/SdJwtDefinition.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/SdJwtDefinition.kt
@@ -146,12 +146,10 @@ fun DisclosableDefObject<String, AttributeMetadata>.findElement(
     }
 
     val head = claimPath.head()
-    require(head is ClaimPathElement.Claim) {
-        "the first element of ClaimPath must be a Claim"
-    }
     val tail = claimPath.tail()?.value.orEmpty()
 
-    return findElement(head, tail)
+    return if (head is ClaimPathElement.Claim) findElement(head, tail)
+    else null
 }
 
 private fun DisclosableDefObject<String, AttributeMetadata>.findElement(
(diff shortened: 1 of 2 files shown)


src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/SdJwtDefinition.kt
```diff
@@ -108,3 +108,95 @@ data class AttributeMetadata(
...
+ * Gets the [AttributeMetadata] of a [DisclosableDef]
+ * @receiver the attribute definition to query
+ * @return the attribute metadata
+ */
+fun DisclosableDef<String, AttributeMetadata>.attributeMetadata(): AttributeMetadata =
+    when (this) {
+        is DisclosableDef.Id<String, AttributeMetadata> -> value
+        is DisclosableDef.Arr<String, AttributeMetadata> -> {
+            check(value is SdJwtArrayDefinition)
+            value.metadata
```
dzarras, line 126: @babisRoutis No need for this duplicate.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/SdJwtDefinition.kt
```diff
@@ -108,3 +108,95 @@ data class AttributeMetadata(
...
+                    def.value.findElement(tail)
+                }
+
+                is DisclosableDef.Id<String, AttributeMetadata> -> null
+            }
+        }
+    }
+}
+
+private fun DisclosableDefArray<String, AttributeMetadata>.findElement(
```
dzarras, line 183: @babisRoutis Why is this private while the one for `DisclosableDefObject` public?

src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/SdJwtDefinition.kt
```diff
@@ -108,3 +108,95 @@ data class AttributeMetadata(
...
+        val head = claimPath.first()
+        val tail = claimPath.drop(1)
+        content[head.name]?.let { element ->
+            if (tail.isEmpty()) element
+            else when (val def = element.value) {
+                is DisclosableDef.Obj<String, AttributeMetadata> -> {
+                    def.value.findElement(tail)
+                }
+
+                is DisclosableDef.Arr<String, AttributeMetadata> -> {
```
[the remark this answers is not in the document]
babisRoutis: I don't follow.
What is the problem that you see?

src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/SdJwtDefinition.kt
```diff
@@ -108,3 +108,95 @@ data class AttributeMetadata(
...
+        }
+    }
+}
+
+private fun DisclosableDefArray<String, AttributeMetadata>.findElement(
+    claimPath: List<ClaimPathElement.Claim>,
+): SdJwtElementDefinition? {
+    return if (claimPath.isEmpty()) null
+    else {
+        val tail = claimPath.drop(1)
```
dzarras, line 188: @babisRoutis I don't think this method is properly defined. Why drop the `head`?

For example, This method is called with the following claim path: ["foo", "bar"], `tail` will be just ["bar"].
1. If the array element definition is an object, shouldn't we call  `def.value.findElement()` using `claimPath` instead of just `tail`?
2. If the array element definition is an array, shouldn't the first element of `claimPath` be `ClaimPathElement.AllArrayElements`?

dzarras (approved):

PR #376, opened 2025-06-20
Introduce `TypeMetadataPolicy`

dzarras:
Currently when Type Metadata resolution is enabled, a Verifier has no way of defining a policy regarding Type Metadata. 

This PR introduces `TypeMetadataPolicy` as a means to rectify the above.
When constructing an `SdJwtVcVerifier`, a Verifier must now also defined a policy regarding Type Metadata.

This PR introduces the following policies:
- `TypeMetadataPolicy.NotUsed`: Type Metadata are not used.
- `TypeMetadataPolicy.Optional`: Type Metadata are optional. In case of resolution success, extra validation are performed against an SD-JWT VC using the resolved Type Metadata. In case of resolution failure, these extra checks are skipped.
- `TypeMetadataPolicy.AlwaysRequired`: Type Metadata are always required. In case of resolution success, extra validation are performed against an SD-JWT VC using the resolved Type Metadata. In case of resolution failure, the SD-JWT VC is rejected.
-  `TypeMetadataPolicy.RequiredFor`: Applies `TypeMetadataPolicy.AlwaysRequired` for a set of specified Vcts, and  `TypeMetadataPolicy.Optional` for everything else.

### DIFF
--- a/Module.md
+++ b/Module.md
@@ -15,14 +15,15 @@ Library's SD-JWT DSL leverages the DSL provided by
 
 - [Issuance](#issuance): As an Issuer use the library to issue a SD-JWT
 - [Holder Verification](#holder-verification): As Holder verify a SD-JWT issued by an Issuer
-- [Presentation Verification](#presentation-verification): As a Verifier verify SD-JWT in simple or in Enveloped Format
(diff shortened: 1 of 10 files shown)


babisRoutis (dismissed):
@dzarras  LGTM

Please check if README.md and dsl.md need changes

PR #389, opened 2025-08-28
Add support for checking Integrity when fetching TypeMetadata or Json Schemas

### LINKED ISSUE #388: Add support for checking Integrity when fetching TypeMetadata or Json Schemas
Currently neither `LookupTypeMetadata` nor `LookupJsonSchema` check the integrity of the documents they fetch.

We must update these interfaces and introduce in their respective Ktor implementation the necessary checks.

To that end:

1. Rename `DocumentIntegrity` value class to `DocumentIntegrities` to better reflect that this can contain multiple integrity values.

2. Introduce:
```kotlin
enum class IntegrityAlgorithm(val name: String) { ... }

data class DocumentIntegrity(val algorithm: IntegrityAlgorithm, val value: String)
```

3. Introduce a method that parses `DocumentIntegrities` to `Set<DocumentIntegrity>`

4. Add a second parameter to the `invoke` methods of both `LookupTypeMetadata`, and `LookupJsonSchema`: `integrity: Set<DocumentIntegrity>?`

5. Update Ktor implementations as needed. 

References:
- https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-08.html#document-integrity
- https://www.w3.org/TR/sri/

atsant:
Closes #388 
Note: targeting release/0.16.0


src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
```diff
@@ -15,23 +15,28 @@
...
+import java.io.ByteArrayInputStream
+import kotlin.io.encoding.Base64 as Base64Kotlin
 
 /**
  * Lookup the Type Metadata of a VCT.
  */
 fun interface LookupTypeMetadata {
 
-    suspend operator fun invoke(vct: Vct): Result<SdJwtVcTypeMetadata?>
+    suspend operator fun invoke(vct: Vct, documentIntegrities: List<DocumentIntegrity>?): Result<SdJwtVcTypeMetadata?>
```
dzarras, line 32: @atsant Let's rename `documentIntegrities` to `expectedIntegrity`.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
```diff
@@ -15,23 +15,28 @@
...
 
     companion object {
         fun firstNotNullOfOrNull(first: LookupTypeMetadata, vararg remaining: LookupTypeMetadata): LookupTypeMetadata {
             val lookups = listOf(first, *remaining)
-            return LookupTypeMetadata { vct ->
+            return LookupTypeMetadata { vct, integrities ->
```
dzarras, line 37: @atsant Here as well. Rename the `integrities` lambda parameter to `expectedIntegrity`.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
```diff
@@ -15,23 +15,28 @@
...
 package eu.europa.ec.eudi.sdjwt.vc
 
+import eu.europa.ec.eudi.sdjwt.platform
 import eu.europa.ec.eudi.sdjwt.runCatchingCancellable
 import io.ktor.client.request.*
 import io.ktor.http.*
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.decodeFromStream
+import java.io.ByteArrayInputStream
+import kotlin.io.encoding.Base64 as Base64Kotlin
```
dzarras, line 25: @atsant Let's drop the `Base64Kotlin` import alias. I don't think we have another class/object named `Base64` in this file.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
```diff
@@ -44,30 +49,57 @@ fun interface LookupTypeMetadata {
...
+    )
+}
+
 /**
  * Lookup a [JsonSchema] from a Uri.
  */
 fun interface LookupJsonSchema {
 
-    suspend operator fun invoke(uri: String): Result<JsonSchema?>
+    suspend operator fun invoke(uri: String, schemaIntegrity: List<DocumentIntegrity>?): Result<JsonSchema?>
```
dzarras, line 95: @atsant Similarly here, rename `schemaIntegrity` to `expectedIntegrity`.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
```diff
@@ -81,11 +113,21 @@ class LookupJsonSchemaUsingKtor(
     private val httpClientFactory: KtorHttpClientFactory = DefaultHttpClientFactory,
 ) : LookupJsonSchema {
 
-    override suspend fun invoke(uri: String): Result<JsonSchema?> {
+    override suspend fun invoke(uri: String, schemaIntegrity: List<DocumentIntegrity>?): Result<JsonSchema?> {
```
dzarras, line 116: @atsant Here as well. `expectedIntegrity`.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/DocumentIntegrityTest.kt
```diff
@@ -0,0 +1,76 @@
...
+import kotlin.test.Test
+
+class DocumentIntegrityTest {
+
+    @Test
+    fun `ensure single valid DocumentIntegrities is being converted correctly to DocumentIntegrity lists`() {
+        val singleValid = DocumentIntegrities("sha384-Li9vy3DqF8tnTXuiaAJuML3ky+er10rcgNR/VqsVpcw+ThHmYcwiB1pbOxEbzJr7")
+        val singleValidIntegrity = singleValid.toDocumentIntegrity()
+
+        assert(singleValidIntegrity.size == 1)
```
dzarras, line 30: @atsant Please do not use Java assertions. Replace with `assertEquals` instead.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/DocumentIntegrityTest.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    @Test
+    fun `ensure multiple valid DocumentIntegrities is being converted correctly to DocumentIntegrity lists`() {
+        val multipleValid = DocumentIntegrities(
+            "sha384-H8BRh8j48O9oYatfu5AZzq6A9RINhZO5H16dQZngK7T62em8MUt1FLm52t+eX6xO " +
+                "sha512-Q2bFTOhEALkN8hOms2FKTDLy7eugP2zFZ1T8LCvX42Fp3WoNr3bjZSAHeOsHrbV1Fu9/A0EzCinRE7Af1ofPrw==",
+        )
+        val multipleValidIntegrity = multipleValid.toDocumentIntegrity()
+
+        assert(multipleValidIntegrity.size == 2)
```
dzarras, line 41: @atsant Same here. Use `assertEquals`.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/DocumentIntegrityTest.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    @Test
+    fun `ensure when a known and one unknown algorithm exists, keep only the known one`() {
+        val knownAndUnknownAlgorithm = DocumentIntegrities(
+            "sha484-H8BRh8j48O9oYatfu5AZzq6A9RINhZO5H16dQZngK7T62em8MUt1FLm52t+eX6xO " +
+                "sha512-Q2bFTOhEALkN8hOms2FKTDLy7eugP2zFZ1T8LCvX42Fp3WoNr3bjZSAHeOsHrbV1Fu9/A0EzCinRE7Af1ofPrw==",
+        )
+        val knownAndUnknownAlgorithmIntegrity = knownAndUnknownAlgorithm.toDocumentIntegrity()
+
+        assert(knownAndUnknownAlgorithmIntegrity.size == 1)
```
dzarras, line 59: @atsant 

1. Replace with `assertEquals`
2. Please verify that indeed the correct `DocumentIntegrity` is kept in the list.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/DocumentIntegrityTest.kt
```diff
@@ -0,0 +1,76 @@
...
+ * limitations under the License.
+ */
+package eu.europa.ec.eudi.sdjwt
+
+import eu.europa.ec.eudi.sdjwt.vc.DocumentIntegrities
+import eu.europa.ec.eudi.sdjwt.vc.toDocumentIntegrity
+import org.junit.jupiter.api.assertThrows
+import kotlin.test.Test
+
+class DocumentIntegrityTest {
```
dzarras, line 23: @atsant Since we are mostly checking an extension function of `DocumentIntegrities`, better rename the file and test class to `DocumentIntegritiesTest`.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/DocumentIntegrityTest.kt
```diff
@@ -0,0 +1,76 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package eu.europa.ec.eudi.sdjwt
+
+import eu.europa.ec.eudi.sdjwt.vc.DocumentIntegrities
+import eu.europa.ec.eudi.sdjwt.vc.toDocumentIntegrity
+import org.junit.jupiter.api.assertThrows
```
dzarras, line 20: @atsant Whenever possible use assertion methods from the `kotlin.test` package. I believe there should be an assertion method called `assertFailsWith`.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/TypeMetadata.kt
```diff
@@ -396,7 +397,38 @@ data class LogoMetadata(
...
 @JvmInline
-value class DocumentIntegrity(val value: String)
+value class DocumentIntegrities(val value: String)
+
+data class DocumentIntegrity(
+    val hashAlgorithm: IntegrityAlgorithm,
+    val value: String,
+)
+
+fun DocumentIntegrities.toDocumentIntegrity(): List<DocumentIntegrity> =
```
dzarras, line 407: @atsant Please mark this `private` we don't have to expose this.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/TypeMetadata.kt
```diff
@@ -396,7 +397,38 @@ data class LogoMetadata(
 // TODO Check this
 @Serializable
 @JvmInline
-value class DocumentIntegrity(val value: String)
+value class DocumentIntegrities(val value: String)
```
dzarras, line 400: @atsant Could you add an `init` block and require that `value` is not blank? I don't think it makes sense to have a `DocumentIntegrities` with a blank value.

dzarras:
@babisRoutis Could you also have a quick look?

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/TypeMetadata.kt
dzarras: Then we either need to move it, or mark it `internal`.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
```diff
@@ -44,30 +49,57 @@ fun interface LookupTypeMetadata {
...
             }
         }
     }
 }
 
+private fun validateIntegrity(expectedIntegrity: List<DocumentHash>, document: ByteArray) {
+    val base64Padding = Base64.withPadding(Base64.PaddingOption.PRESENT)
+
+    require(
+        expectedIntegrity.any {
```
babisRoutis, line 76: I am not sure that this is correct.


The SRI specification (https://www.w3.org/TR/sri/#agility) has the following 

> the user agent will choose the strongest hash function in the list, and use that metadata to validate the response (as described below in the § 3.3.2 Parse metadata and § 3.3.3 Get the strongest metadata from set algorithms).
 
 
I believe that  `validateIntegrity` should become a public class (ie SRIntegrityValidator), that has a set of hash  functions that the (library) caller considers goods.

This validator should should peek the strongest alg that is allowed to use and it is present in provided the `expectedHashes`

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
```diff
@@ -81,11 +113,21 @@ class LookupJsonSchemaUsingKtor(
     private val httpClientFactory: KtorHttpClientFactory = DefaultHttpClientFactory,
 ) : LookupJsonSchema {
 
-    override suspend fun invoke(uri: String): Result<JsonSchema?> {
+    override suspend fun invoke(uri: String, expectedIntegrity: List<DocumentHash>?): Result<JsonSchema?> {
         val url = runCatchingCancellable { Url(uri) }.getOrNull()
```
babisRoutis, line 117: Use runCatching in this line. The constructor of `Url` doesn't throw CancellableException

src/test/kotlin/eu/europa/ec/eudi/sdjwt/DocumentIntegrityTest.kt
```diff
@@ -0,0 +1,76 @@
...
+        assertFailsWith<IllegalArgumentException> {
+            DocumentIntegrity(
+                "sha484-H8BRh8j48O9oYatfu5AZzq6A9RINhZO5H16dQZngK7T62em8MUt1FLm52t+eX6xO " +
+                    "sha512-Q2bFTOhEALkN8hOms2FKTDLy7eugP2zFZ1T8LCvX42Fp3WoNr3bjZSAHeOsHrbV1Fu9/A0EzCinRE7Af1ofPrw==",
+            )
+        }
+    }
+
+    @Test
+    fun `ensure that when options are present, they are removed and ignored`() {
```
dzarras, line 64: @atsant Update the test case name. The options are indeed from the `encodedHash` but are not per-se ignore. They are used to populate a different property of `DocumentHash`.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SRIValidatorTest.kt
```diff
@@ -0,0 +1,78 @@
...
+import eu.europa.ec.eudi.sdjwt.vc.SRIValidator
+import kotlin.test.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class SRIValidatorTest {
+
+    @Test
+    fun `ensure that hash of content to validate exists in strongest provided hashes`() {
+        val singleValid = DocumentIntegrity(
```
dzarras, line 28: @atsant Please add some comments in the code that indicate that the valid hash of the test string is one of the sha512 hashes.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SRIValidatorTest.kt
```diff
@@ -0,0 +1,78 @@
...
+                "sha512-tLL38NKkjSrUzPZcxdw2Cje4pvsXFicllTGy7hgenGSdRfaU7jSVqscGaV9OjUq6UmeHJXyoPYrCiwQcR3r5uw==",
+
+        )
+        val validationResult = SRIValidator().isValid(singleValid, "asdasdas".toByteArray(Charsets.UTF_8))
+
+        assertTrue(validationResult)
+    }
+
+    @Test
+    fun `hash of content to validate does not exists in strongest provided hashes`() {
```
dzarras, line 41: @atsant Please update the name of the test case. For instance `Validation fails when ...`

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SRIValidatorTest.kt
```diff
@@ -0,0 +1,78 @@
...
+        val singleValid = DocumentIntegrity(
```
dzarras, line 42: @atsant Please rename the variable. In this context none of the provided hashes match the test string.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SRIValidatorTest.kt
```diff
@@ -0,0 +1,78 @@
...
+                "sha384-tLL38NKkjSrUzPZcxdw2Cje4pvsXFicllTGy7hgenGSdRfaU7jSVqscGaV9OjUq6UmeHJXyoPYrCiwQcR3r5uw==",
+
+        )
+        val validationResult = SRIValidator().isValid(singleValid, "asdasdas".toByteArray(Charsets.UTF_8))
+
+        assertFalse(validationResult)
+    }
+
+    @Test
+    fun `hash of content to validate does not exists in provided hashes`() {
```
dzarras, line 55: @atsant Please rename the test case. `Validation fails when ...`.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SRIValidatorTest.kt
```diff
@@ -0,0 +1,78 @@
...
+        val singleValid = DocumentIntegrity(
```
dzarras, line 56: @atsant Same here, use a better name for the variable.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SRIValidatorTest.kt
```diff
@@ -0,0 +1,78 @@
...
+                "sha384-H8BRh8j48O9oYatfu5AZzq6A9RINhZO5H16dQZngK7T62em8MUt1FLm52t+eX6xO?asdasdsadsadsad " +
+                "sha512-Q2bFTOhEALkN8hOms2FKTDLy7eugP2zFZ1T8LCvX42Fp3WoNr3bjZSAHeOsHrbV1Fu9/A0EzCinRE7Af1ofPrw== ",
+        )
+        val validationResult = SRIValidator().isValid(singleValid, "asdasdas".toByteArray(Charsets.UTF_8))
+
+        assertFalse(validationResult)
+    }
+
+    @Test
+    fun `hash of content is hash SHA-512 but exists as SHA-386 in provided hashes`() {
```
dzarras, line 67: @atsant I don't think this test case is needed. If as sha512 hash is marked as sha384, then by definition it will never match. It's like having a sha384 hash with an invalid value.

atsant:
@babisRoutis Can you have a look again please ?

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/GetSubResourceKtorOps.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+package eu.europa.ec.eudi.sdjwt.vc
+
+import io.ktor.http.Url
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.decodeFromStream
+import java.io.ByteArrayInputStream
+
+internal class GetSubResourceKtorOps(
+    private val httpClientFactory: KtorHttpClientFactory,
```
babisRoutis, line 24: Please remove the `httpClientFactory`.


Instead define `HttpClient.getOrNull()`


With this change, you can use it  as

```kotlin

with(GetSubResourceKtorOps(sriValidator)){
   httpClient.getOrNul(....)
}


```

babisRoutis (approved):

PR #456, opened 2026-03-04
SD-JWT verification: Missing Validity Checks for claims ‘nbf’, and ‘exp’.

### LINKED ISSUE #451: SD-JWT verification: Missing Validity Checks for claims ‘nbf’, ‘exp’ and ‘aud’
The implementation for `SdJwtVerifier` is currently missing checks for:

* `nbf`
* `exp`
* `aud`

Even though a `JwtSignatureVerifier` can be instantiated from a `JWTProcessor` that uses a `JWTClaimSetVerifier`, that's not always the case.

We must update the implementation of `SdJwtVerifier` to also perform these validation checks.

We could introduce an extra parameter to the `verify` methods, i.e. `VerificationContext` as follows:
```kotlin
data class VerificationContext(
   val notBefore: Instant? = null,
   val expiresAt: Instant? = null,
   val audience: String? = null,
)
```

Where the presence of a property means the callers requires the presence and validity of each respective claim.

Given the above, the `doVerify` method in `SdJwtVerifier.kt` can be updated to include the missing checks.

One extra consideration is the fact that the interface `SdJwtVerifier` is not sealed, meaning one can provide an implementation that skips these checks all-together. 
We should consider whether sealing this interface is also desirable.

atsant:
Closes #451

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
@@ -586,10 +586,10 @@ data class ValidityVerificationContext private constructor(
             expiresAt = expiresAt?.dropNanoSeconds(),
             audience = audience,
         )
-
-        /**
-         * Returns this [Instant] truncated to millisecond precision.
-         */
-        private fun Instant.dropNanoSeconds(): Instant = Instant.fromEpochMilliseconds(toEpochMilliseconds())
     }
 }
+
+/**
+ * Returns this [Instant] truncated to millisecond precision.
+ */
+private fun Instant.dropNanoSeconds(): Instant = Instant.fromEpochMilliseconds(toEpochMilliseconds())
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
@@ -486,12 +486,12 @@ private fun JsonObject.exp(): Instant? =
         }
 
 /**
- * Returns this [Instant] truncated to millisecond precision.
+ * Returns a new [Instant] that has seconds precision.
  */
-private fun Instant.dropNanoSeconds(): Instant = Instant.fromEpochMilliseconds(toEpochMilliseconds())
+private fun Instant.withSecondsPrecision(): Instant = Instant.fromEpochSeconds(epochSeconds, 0)
 
 private fun JsonObject.validate(clock: Clock, skew: Duration) {
-    val now = clock.now().dropNanoSeconds()
+    val now = clock.now().withSecondsPrecision()
     val nbf = nbf()
     if (null != nbf) {
         if ((nbf - skew) > now) {
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
@@ -455,50 +455,8 @@ private suspend fun <JWT> doVerify(
     val disclosures = toDisclosures(unverifiedDisclosures)
     val (recreated, _) = SdJwtRecreateClaimsOps.recreateClaimsAndDisclosuresPerClaim(jwtClaims, disclosures).getOrThrow()
 
-    val nbf = recreated[RFC7519.NOT_BEFORE]?.jsonPrimitive?.content?.let {
-        Instant.fromEpochSeconds(it.toLong())
-    }
-    val exp = recreated[RFC7519.EXPIRATION_TIME]?.jsonPrimitive?.content?.let {
-        Instant.fromEpochSeconds(it.toLong())
-    }
-    val aud: List<String>? = recreated[RFC7519.AUDIENCE]?.let { element ->
-        when (element) {
-            is JsonArray ->
-                element.map { it.jsonPrimitive.content }
-            is JsonPrimitive ->
-                element.contentOrNull?.let(::listOf)
-            else -> null
-        }
-    }
-
-    validityVerificationContext.let { context ->
-        fun invalidJwt(message: String): Result<Nothing> =
-            Result.failure(InvalidJwt(message).asException())
-
-        context.notBefore?.let { ctxNotBefore ->
-            nbf?.let { tokenNotBefore ->
-                if (tokenNotBefore > ctxNotBefore) {
-                    return invalidJwt("JWT not valid before $ctxNotBefore")
-                }
-            }
-        }
-
-        context.expiresAt?.let { ctxExpiresAt ->
-            exp?.let { tokenExpiresAt ->
-                if (tokenExpiresAt < ctxExpiresAt) {
-                    return invalidJwt("JWT not valid since expired at $ctxExpiresAt")
-                }
-            }
-        }
-
-        context.audience?.let { expectedAudience ->
-            aud?.let { tokenAudiences ->
-                if (expectedAudience in tokenAudiences) {
-                    return invalidJwt("JWT not having valid aud value $expectedAudience")
-                }
-            }
-        }
-    }
+    // verified validity of nbf, exp and aud claims if exists
+    validateValidity(recreated, validityVerificationContext).getOrThrow()
 
     // Check Key binding
     val expectedDigest = SdJwtDigest.digest(hashAlgorithm, unverifiedSdJwt).getOrThrow()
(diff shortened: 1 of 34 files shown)


src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/NimbusSdJwtVcVerifierFactory.kt
```diff
@@ -47,6 +47,7 @@ internal object NimbusSdJwtVcVerifierFactory : SdJwtVcVerifierFactory<NimbusSign
     override fun invoke(
         issuerVerificationMethod: IssuerVerificationMethod<SignedJWT, JWK, List<X509Certificate>>,
         typeMetadataPolicy: TypeMetadataPolicy,
+        validityVerificationContext: ValidityVerificationContext,
```
dzarras: @atsant This must not be a parameter in `SdJwtVcVerifierFactory`.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/NimbusSdJwtVcVerifierFactory.kt
```diff
@@ -76,7 +78,7 @@ private class NimbusSdJwtVcVerifier(
 
     override suspend fun verify(unverifiedSdJwt: String): Result<SdJwt<NimbusSignedJWT>> =
```
dzarras, line 77: @atsant `validityVerificationContext` must a parameter in `SdJwtVcVerifier.verify`

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -448,6 +470,54 @@ private suspend fun <JWT> doVerify(
...
+        when {
+            element is JsonArray ->
+                element.map { it.jsonPrimitive.content }
+            element is JsonPrimitive && element.isString ->
+                listOf(element.content)
+            else -> error("${RFC7519.AUDIENCE} can be JsonArray or JsonPrimitive with string content")
+        }
+    }
+
+    validityVerificationContext.let { context ->
```
dzarras: @atsant No need to use `let` here

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -448,6 +470,54 @@ private suspend fun <JWT> doVerify(
     sdJwt to kbJwt
 }
 
+private fun validateValidity(recreated: JsonObject, validityVerificationContext: ValidityVerificationContext): Result<Unit> {
```
dzarras: @atsant No need to return `Result` have the function throw the `InvalidJwt` exception instead.

dzarras: @atsant Try to define `validateValidity` as an extension function over `ValidityVerificationContext` instead like so:

`ValidityVerificationContext.validate(payload: JsonObject)`

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -448,6 +470,48 @@ private suspend fun <JWT> doVerify(
...
+        when {
+            element is JsonArray ->
+                element.map { it.jsonPrimitive.content }
+            element is JsonPrimitive && element.isString ->
+                listOf(element.content)
+            else -> error("${RFC7519.AUDIENCE} can be JsonArray or JsonPrimitive with string content")
+        }
+    }
+
+    this.notBefore?.let { notBeforeContext ->
```
dzarras: @atsant You could simplify this with a simple if check
```kotlin
if (null != this.notBefore && null != nbf) { ... }
```

dzarras: This applies to the checks below as well.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -472,3 +536,25 @@ internal fun JwsJsonSupport.parseIntoStandardForm(unverifiedSdJwt: JsonObject):
     val kbJwtSerialized = unverifiedKBJwt ?: ""
     return "$jwtAndDisclosures$kbJwtSerialized"
 }
+
+data class ValidityVerificationContext(
```
dzarras: @atsant Please document this class, and what each property checks. Additionally also document the factory method in the companion object, and what the extra skew parameter does.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -472,3 +536,25 @@ internal fun JwsJsonSupport.parseIntoStandardForm(unverifiedSdJwt: JsonObject):
...
+) {
+
+    companion object {
+        operator fun invoke(
+            notBefore: Instant? = null,
+            expiresAt: Instant? = null,
+            audience: String? = null,
+            skew: Duration,
+        ): ValidityVerificationContext {
+            require(skew >= Duration.ZERO) { "skew must be positive" }
```
dzarras: @atsant the message doesn't reflect the check. You require skew to be positive or zero.

Better change this to 
```kotlin
require(!skew.isNegative) { "skew cannot be negative" }
```

dzarras (changes requested):

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
```diff
@@ -47,53 +48,69 @@ interface SdJwtVcVerifier<out JWT> {
      * @param unverifiedSdJwt the SD-JWT to be verified.
      * A JSON Object that is expected to be in general
      * or flatten form as defined in RFC7515 and extended by SD-JWT specification.
+     * @param validityVerificationContext the context used to verify the validity of the SD-JWT of claims `nbf`, `exp` and/or `aud`
```
dzarras: @atsant Please update the description to "context used to verify the validity of the issuer-signed JWT claims `nbf`, `exp`, and `aud`" wherever needed.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -472,3 +524,72 @@ internal fun JwsJsonSupport.parseIntoStandardForm(unverifiedSdJwt: JsonObject):
...
+        ): ValidityVerificationContext = ValidityVerificationContext(
+            notBefore = notBefore?.dropNanoSeconds(),
+            expiresAt = expiresAt?.dropNanoSeconds(),
+            audience = audience,
+        )
+
+        /**
+         * Returns this [Instant] truncated to millisecond precision.
+         */
+        private fun Instant.dropNanoSeconds(): Instant = Instant.fromEpochMilliseconds(toEpochMilliseconds())
```
dzarras: @atsant make the extension function private to the file instead of the companion object.

src/test/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifierVerificationOfClaimsTest.kt
```diff
@@ -0,0 +1,219 @@
...
+
+import eu.europa.ec.eudi.sdjwt.DefaultSdJwtOps.NoSignatureValidation
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlin.test.fail
+import kotlin.time.Instant
+
+class SdJwtVerifierVerificationOfClaimsTest {
```
dzarras: @atsant Please rename the test class to `SdJwtVerifierTest` since we are testing the behavior of this class.

dzarras (dismissed):

dzarras:
@babisRoutis What do you think of these changes?

dzarras:
@babisRoutis Could please check the reworked PR? 

Currently we are not validating the contents of the `aud` claim. We should consider whether it is in scope for the library or not.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -448,6 +460,53 @@ private suspend fun <JWT> doVerify(
...
+            Instant.fromEpochSeconds(it.content.toLong())
+        }
+
+/**
+ * Returns a new [Instant] that has seconds precision.
+ */
+private fun Instant.withSecondsPrecision(): Instant = Instant.fromEpochSeconds(epochSeconds, 0)
+
+private fun JsonObject.validate(clock: Clock, skew: Duration) {
+    val now = clock.now().withSecondsPrecision()
```
babisRoutis, line 494: `iat`?

babisRoutis: The answer is no

https://www.rfc-editor.org/rfc/rfc7519#section-4.1.6

babisRoutis (approved):

PR #457, opened 2026-03-09
Make KeyBinding JWT checks more fine grained

### LINKED ISSUE #452: SD-JWT+KB Verification: Make challenge more fine grained
`SdJwtVcVerifier` allows a verifier to provide a challenge, i.e. a claim set, that must be present and exactly matched in the Key-Binding JWT.

This can cause issues though with the time-based `iat` claim.
We should make the Key-Binding JWT challenge more fine grained and allow a caller to provide a clock skew as well.

For instance Challenge could be expressed as follows:
```kotlin
data class Challenge(
   val audience: String,
   val nonce: String,
   val sdHash: String,
   val issuedAt: Instant,
   val clockSkew: Duration,
)
```

Additionally, it might be worth considering sealing the interfaces `SdJwtVcVerifier`, and disallow clients from using custom implementations as a meant to bypass certain checks.

dzarras:
This PR:

Updates KeyBinding JWT challenge from a plain `JsonObject` to a data class that contains:
1. a `ClosedRange` within which `iat` is acceptable
2. other claims that must be exactly matched withing the Key-Binding JWT claimset, encoded as a `JsonObject`

The only publicly exposed constructor requires `iat`, an optional clock `skew` to define a `ClosedRange` for `iat`, `audience`, `nonce`, and an optional extension function over a `JsonObjectBuilder` that can be used to include other claims, such as Transaction Data Hashes, that must be matched exactly withing the Key-Binding JWT claimset.

Additionally, the `KeyBindingJwtChallenge` parameter has been added to `SdJwtVerifier` methods that verify SD-JWT+KB. Previously callers could not provide a challenge when verifying an SD-JWT+KB using `SdJwtVerifier`.

Finally, an extra value class was added, to parse and verify the claimset of Key-Binding JWT. 
The factory method of `KeyBindingJwtClaims` verifies that:
1. `iat` is present, and is a long
2. `aud` is present, and is a string (RFC9901: doesn't allow `aud` to be an array of strings in Key-Binding JWT)
3. `nonce` is present, and is a string
4. `sd_hash` is present, and is a string

Previously we only verified that `iat`, `aud`, `nonce,` and `sd_hash` where present in the Key-Binding JWT without verifying their types.

Closes #452 
Closes #454

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/NimbusIntegration.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/NimbusIntegration.kt
@@ -74,9 +74,12 @@ import com.nimbusds.jwt.proc.JWTProcessor as NimbusJWTProcessor
  * - The header contains typ claim equal to `kb+jwt`
  * - The header contains the signing algorithm claim
  * - The JWT must be signed by the given [holderPubKey]
- * - Claims `iat`, `aud`, `nonce`, and `sd_hash` must be present to the JWT payload
+ * - If the challenge is provided it should present in JWT payload
+ * - Claims `aud`, `iat` and `nonce`must be present to the JWT payload
  *
  * @param holderPubKey the public key of the holder
+ * @param challenge an optional challenge provided by the verifier, to be signed by the holder as the Key-binding JWT.
+ * If provided, Key Binding JWT payload should contain the challenge as is.
  * @return
  */
 internal fun <PubKey> keyBindingJWTProcess(
@@ -162,8 +165,10 @@ object NimbusSdJwtOps :
      *
      *  @param holderPubKeyExtractor a function that extracts the holder's public key from the payload of the SD-JWT.
      * If not provided, it is assumed that the SD-JWT issuer used the confirmation claim (see [cnf]) for this purpose.
+
      * @param challenge an optional challenge provided by the verifier, to be signed by the holder as the Key-binding JWT.
      * If provided, Key Binding JWT payload should contain the challenge as is.
+
      * @see keyBindingJWTProcess
      */
     fun KeyBindingVerifier.Companion.mustBePresentAndValid(
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/NimbusIntegration.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/NimbusIntegration.kt
@@ -79,12 +79,14 @@ import com.nimbusds.jwt.proc.JWTProcessor as NimbusJWTProcessor
  * @param holderPubKey the public key of the holder
  * @return
  */
-internal fun <PubKey> keyBindingJWTProcess(holderPubKey: PubKey):
-    NimbusJWTProcessor<NimbusSecurityContext> where PubKey : NimbusJWK, PubKey : NimbusAsymmetricJWK =
+internal fun <PubKey> keyBindingJWTProcess(
+    holderPubKey: PubKey,
+    challenge: NimbusJWTClaimsSet? = null,
+): NimbusJWTProcessor<NimbusSecurityContext> where PubKey : NimbusJWK, PubKey : NimbusAsymmetricJWK =
     JwkSourceJWTProcessor(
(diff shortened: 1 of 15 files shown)


dzarras:
@babisRoutis Please check when time allows.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -427,12 +515,62 @@ interface SdJwtVerifier<JWT> {
     }
 }
 
+/**
+ * Challenge for Key-Binding JWT.
+ *
+ * @property issuedAt window within which `iat` of the Key-Binding JWT is considered valid
+ * @property exactMatchClaims other claims that must be exactly matched in the Key-Binding JWT claims-set
+ */
+data class KeyBindingJwtChallenge private constructor(
```
babisRoutis: Name `KeyBindingJwtChallenge` is confusing, IMO

The name suggests the content of the Challenge should to the Wallet
The contents of the class are restrictions to validate a KeyBindingJwt.

Consider a name like `ChallengePredicate`

src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt
```diff
@@ -427,12 +515,62 @@ interface SdJwtVerifier<JWT> {
...
+         * @param audience expected `aud` of the Key-Binding JWT
+         * @param nonce expected `nonce` of the Key-Binding JWT
+         * @param skew **positive** duration, if provided defines a time window with [iat] within which `iat` of Key-Binding JWT is considered valid
+         * @param exactMatchClaimsBuilder builder for other claims that must be exactly matched withing the Key-Binding JWT claims-set
+         */
+        operator fun invoke(
+            issuedAt: Instant,
+            audience: String,
+            nonce: String,
+            skew: Duration? = null,
```
babisRoutis: skew shouldn't be nullable. Just use a sensible default of 5 secs

src/main/kotlin/eu/europa/ec/eudi/sdjwt/NimbusIntegration.kt
```diff
@@ -90,7 +90,7 @@ internal fun <PubKey> keyBindingJWTProcess(
         NimbusDefaultJOSEObjectTypeVerifier(NimbusJOSEObjectType("kb+jwt")),
         NimbusDefaultJWTClaimsVerifier(
             challenge ?: NimbusJWTClaimsSet.Builder().build(),
-            setOf(RFC7519.AUDIENCE, RFC7519.ISSUED_AT, "nonce"),
+            setOf(RFC7519.ISSUED_AT, RFC7519.AUDIENCE, RFC9901.CLAIM_NONCE, RFC9901.CLAIM_SD_HASH),
```
babisRoutis, line 93: Here we require `aud`. Yet, the `KeyBindingJwtChallenge` invoke method doesn't include and `aud` array.

babisRoutis (changes requested):
Please check my comments.
Most of them, minor.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/NimbusIntegration.kt
dzarras: `ChallengePredicate.Companion.invoke()` contains an `audience` parameter, which is used to match the `aud` claim in the Key-Binding JWT. 
Check here: https://github.com/niscy-eudiw/eudi-lib-jvm-sdjwt-kt/blob/feat/keybinding-jwt-checks/src/main/kotlin/eu/europa/ec/eudi/sdjwt/SdJwtVerifier.kt#L545

It's a single string and not an array of strings because per RFC9901 `aud` in Key-Binding JWT is a single string.

Reference: https://www.rfc-editor.org/rfc/rfc9901#section-4.3-3.2.2.2

babisRoutis:
@dzarras 
Are there changes for the README?

dzarras:
> @dzarras Are there changes for the README?

No. I didn't spot anything that needed updates. For Key-Binding JWT verification we point the readers to the examples in the test sources.

babisRoutis (approved):

PR #458, opened 2026-03-10
Add Status resolution abstraction

### LINKED ISSUE #453: Allow `status` to be checked in SD-JWT VCs
As part of the validation rules of section 3.4 in https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-13.html#name-verification-and-processing, the status should be checked if present within credentials. 
Currently no such check is performed during verification in `SdJwtVcVerifier`.

We should update `SdJwtVcVerifierFactory` and allow callers to provide a `status` verification mechanism, and update implementations appropriately.

For instance:
```kotlin
sealed interface StatusVerifier {
   data object Ignored : StatusVerifier 

   data class TokenStatusList(val verifier: suspend (StatusListReference) -> StatusVerificationResult) : StatusVerifier 

   data class Other(val verifier: suspend (JsonObject) -> StatusVerificationResult) : StatusVerifier
}
```

It's worth checking whether [Statium](https://github.com/eu-digital-identity-wallet/eudi-lib-kmp-statium) could be brought in as an implementation dependency and provide an implementation for Token Status Lists.

dzarras:
This PR adds a new abstraction `CheckWithTokenStatusList`.

This new abstraction is used to resolve the Status of an SD-JWT VC, during verification, when `status` is present and is a `JsonObject`.

If an `SdJwtVcVerifier` is instantiated with a `CheckWithTokenStatusList` instance, during SD-JWT VC verification if the SD-JWT VC contains a `status` claim and contains a Token Status List reference (`status_list`):
1. The `Status` of the SD-JWT VC is fetched using the provided `CheckWithTokenStatusList` instance
2. We ensure that `Status` is `Valid`

In case `CheckWithTokenStatusList` fails, or `Status` is `NonValid`, the SD-JWT VC is rejected with an `SdJwtVcVerificationError`.

Closes #453

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
@@ -154,6 +154,21 @@ sealed interface SdJwtVcVerificationError {
             }
         }
     }
+
+    /**
+     * Verification error regarding Status resolution and validation.
+     */
+    sealed interface StatusVerificationError : SdJwtVcVerificationError {
+        /**
+         * Status could not be resolved.
+         */
+        class StatusResolutionFailure(val cause: Throwable) : StatusVerificationError
+
+        /**
+         * The Status of an SD-JWT VC is not valid.
+         */
+        class StatusIsNotValid(val invalid: Status.Invalid) : StatusVerificationError
+    }
 }
 
 internal fun raise(error: SdJwtVcVerificationError): Nothing = throw SdJwtVerificationException(VerificationError.SdJwtVcError(error))
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
@@ -156,18 +156,18 @@ sealed interface SdJwtVcVerificationError {
     }
 
     /**
-     * Verification error regarding Status resolution and validation.
+     * Verification error regarding Status check and validation.
      */
     sealed interface StatusVerificationError : SdJwtVcVerificationError {
         /**
-         * Status could not be resolved.
+         * Status could not be checked.
          */
-        class StatusResolutionFailure(val cause: Throwable) : StatusVerificationError
+        class StatusCheckFailure(val message: String, val cause: Throwable) : StatusVerificationError
 
         /**
-         * The Status of an SD-JWT VC is not valid.
+         * The Status of an SD-JWT VC is non-valid.
          */
-        class StatusIsNotValid(val invalid: Status.Invalid) : StatusVerificationError
+        data class NonValidStatus(val status: Status.NonValid) : StatusVerificationError
     }
 }
 
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/Status.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/Status.kt
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2023-2026 European Commission
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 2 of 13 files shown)


src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/Status.kt
```diff
@@ -0,0 +1,33 @@
...
+ */
+sealed interface Status {
+    data object Valid : Status
+    class Invalid(val message: String, val cause: Throwable?) : Status
+}
+
+/**
+ * Resolves the status of an SD-JWT VC given the contents of the `status` claim.
+ */
+fun interface ResolveStatus {
```
babisRoutis: Rename to `CheckWithTokenStatusList`

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/Status.kt
```diff
@@ -0,0 +1,33 @@
...
+package eu.europa.ec.eudi.sdjwt.vc
+
+import kotlinx.serialization.json.JsonObject
+
+/**
+ * The status of an SD-JWT VC.
+ */
+sealed interface Status {
+    data object Valid : Status
+    class Invalid(val message: String, val cause: Throwable?) : Status
```
babisRoutis: Invalid is wrong. 

Status interface, needs to be re-factored

**Option 1**
As sealed interface 
- Valid 
- NonValid(val value: UByte) 

**Option 2**
As enumeration class 
- Valid
- NonValid 

**Option 3**
- Valid
- Invalid
- Suspended
- ApplicationSpecific

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
```diff
@@ -154,6 +154,21 @@ sealed interface SdJwtVcVerificationError {
...
+    sealed interface StatusVerificationError : SdJwtVcVerificationError {
+        /**
+         * Status could not be resolved.
+         */
+        class StatusResolutionFailure(val cause: Throwable) : StatusVerificationError
+
+        /**
+         * The Status of an SD-JWT VC is not valid.
+         */
+        class StatusIsNotValid(val invalid: Status.Invalid) : StatusVerificationError
```
babisRoutis: The `invalid` is not correct. It gives the impression that an error occurred, confusing runtime errors, with status representing not Valid SD-JWT-VC

dzarras: Would refactoring it to a data object with no other properties be better?

i.e.
```kotlin
data object StatusIsNotValid : StatusVerificationError
```

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/Status.kt
dzarras: I'll go with `Option 2`. 

I don't like `Options 1`, and `Option 3`. They seem tightly coupled with Token Status List, and SD-JWT VC doesn't restrict the status mechanism being used.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/Status.kt
dzarras: I don't agree. 
SD-JWT VC doesn't mandate the usage of Token Status List. 
A more generic name should be used. e.g. `CheckStatus`.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/SdJwtVcVerifier.kt
babisRoutis: I would suggest including an error that would indicate a runtime error while trying to check the status (if not specified). But in any case, keep distinct what is expected ( an SD-JWT-VC to be Valid or Not Valid ) and what is not expected (Runtime Error)

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/Status.kt
babisRoutis: The correct abstraction is Option 1, assuming that we are creating a TokenStatusList-based check
You can consider 
Valid as a data object 
Other having a  byte value (for allowing introspection)

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/Status.kt
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package eu.europa.ec.eudi.sdjwt.vc
+
+/**
+ * The status of an SD-JWT VC.
+ */
+sealed interface Status {
+    data object Valid : Status
+    data class NonValid(val value: UByte) : Status
```
vafeini: I would allow caller to provide a message explaining the invalidity. 
An status is valid/invalid in the context of a specific busines logic.  
For example `suspended` status might be valid in specific use cases.
Also status-list spec defines application specific status types that are meaningful per business case. Their validity is case specific.

vafeini (dismissed):

PR #475, opened 2026-05-11
Reject an SD-JWT/SD-JWT VC when a Disclosure contains a claim with a reserved name

### LINKED ISSUE #473: Missing RFC 9901 §7.1 step 3.c.ii validation: forbidden claim names (_sd, ...) not checked during verification
## Summary
During SD-JWT verification, the library does not reject disclosures whose claim name is `_sd` or `...`, as required by [RFC 9901 §7.1 step 3.c.ii](https://www.rfc-editor.org/rfc/rfc9901#section-7.1). It states that when processing a digest found in an object's _sd array, the verifier MUST reject the SD-JWT if:

> If the claim name is `_sd` or `...`, the SD-JWT **MUST** be rejected.

## Current behavior
The reserved-name check in `Disclosure.objectProperty()` (Disclosure.kt) is enforced only at issuance time. During verification, `replaceObjectDigests` in RecreateClaims.kt calls `disclosure.claim()` and uses the resulting name without checking it against `_sd` or `...` A maliciously crafted SD-JWT containing a disclosure with claim name `_sd` or `...` will pass verification without error.

## What is already handled
For completeness, the other two conditions from the same spec step are covered:

Disclosure must be a 3-element array — a 2-element (ArrayElement) disclosure found under an _sd digest is rejected by the check(disclosure is Disclosure.ObjectProperty) guard in `replaceObjectDigests`.

Claim name must not already exist at the _sd level — the disclosuresPerClaimVisitor throws if two disclosures resolve to the same ClaimPath, catching duplicate claim names at the same object level.
Expected behavior

When reconstructing claims from disclosures during verification, if a disclosure resolved from an _sd array has a claim name of _sd or ..., the SD-JWT MUST be rejected.

## Suggested fix
In replaceObjectDigests (RecreateClaims.kt), after val (name, value) = disclosure.claim(), add:

```kotlin
val forbidden = setOf(RFC9901.CLAIM_SD, RFC9901.CLAIM_ARRAY_ELEMENT_DIGEST)  
require(name !in forbidden) {  
    "Disclosure claim name '$name' is forbidden per RFC 9901 §7.1 step 3.c.ii"  
}
```

This mirrors the check already present in Disclosure.ensureValidAttributeName() at issuance time, and brings verification into compliance with the spec.

dzarras:
This PR adds a missing check in `Disclosure.decode()` that ensures that `ObjectProperties` do not use reserved claim names per [RFC9901](https://www.rfc-editor.org/rfc/rfc9901#section-7.1).

Tests have been added to verify the behavior of:
- `SdJwtVerifier`
- `SdJwtVcVerifier`
- `UnverifiedIssuanceFrom`

which are the entry points that accept unvalidated client input.

Fixes #473

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/Disclosure.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/Disclosure.kt
@@ -68,8 +68,15 @@ sealed interface Disclosure {
             val array = Json.decodeFromString<JsonArray>(base64Decoded)
             when (array.size) {
                 3 -> {
+                    fun String.ensureNotReserved() {
+                        val reserved = setOf(RFC9901.CLAIM_SD, RFC9901.CLAIM_ARRAY_ELEMENT_DIGEST)
+                        require(this !in reserved) {
+                            "Disclosed claim cannot be one of: ${reserved.joinToString()}"
+                        }
+                    }
+
                     val salt = array[0].jsonPrimitive.content
-                    val claimName = array[1].jsonPrimitive.content
+                    val claimName = array[1].jsonPrimitive.content.also { it.ensureNotReserved() }
                     val claimValue = array[2]
                     Triple(salt, claimName, claimValue)
                 }
--- a/src/main/kotlin/eu/europa/ec/eudi/sdjwt/Disclosure.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/sdjwt/Disclosure.kt
@@ -68,15 +68,8 @@ sealed interface Disclosure {
             val array = Json.decodeFromString<JsonArray>(base64Decoded)
             when (array.size) {
                 3 -> {
-                    fun String.ensureNotReserved() {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/eu/europa/ec/eudi/sdjwt/Disclosure.kt
```diff
@@ -68,8 +68,15 @@ sealed interface Disclosure {
             val array = Json.decodeFromString<JsonArray>(base64Decoded)
             when (array.size) {
                 3 -> {
+                    fun String.ensureNotReserved() {
+                        val reserved = setOf(RFC9901.CLAIM_SD, RFC9901.CLAIM_ARRAY_ELEMENT_DIGEST)
```
babisRoutis: Please include `_sd_alg`, as well.

dzarras: Done

babisRoutis (approved):

PR #506, opened 2026-08-31
Make `SdJwtSerialization.parseJWSJson()` strict

### LINKED ISSUE #498: Parsing a malformed JWS Json serialized SD-JWT with malformed contents doesn't raise an error
A `disclosures` member that is not an array, or an array with a single non-string element, collapses to an empty list rather than raising an error, and the same happens when the unprotected header is not an object.

When trying to parse a malformed JWS Json serialized SD-JWT the library must properly report the error and not silently fallback to empty lists.

The function must be reworked to not be lenient. For instance if `signatures` is present is must be a `JsonArray`, etc...

dzarras:
Closes #498


dzarras:
@babisRoutis PR updated. Introduced `JwsJsonSupport.tryParseJWSJson()` which returns a `Result`. `JwsJsonSupport.parseJWSJson()` delegates to `JwsJsonSupport.tryParseJWSJson()`. `JwsJsonSupport` is an internal object.

babisRoutis (approved):

PR #508, opened 2026-08-31
Introduce a limit in ResolveTypeMetadata

### LINKED ISSUE #494: Introduce an upper bound to `ResolveTypeMetadata`
Currently `ResolveTypeMetadata` does contain an upper bound that can potentially terminate the resolution.
A host answering each request with a fresh, distinct extends URL keeps the loop running indefinitely.
A Verifier that integrates the library could receive a handful of concurrent presentations and have his connection pool, threads and heap saturated. 

We must introduce a configurable upper bound.

dzarras:
Closes #494 

This PR introduces a limit on the number of parent SD-JWT VC Type Metadata documents that can be resolved.
The resolution was previously unbound, and could lead to resource exhaustion when trying to resolve Type Metadata controlled by a malicious party.


babisRoutis:
@dzarras 

We don't have to introduce a breaking change.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
```diff
@@ -90,8 +89,16 @@ private fun ResolvedTypeMetadata.Companion.empty(vct: Vct): ResolvedTypeMetadata
...
 
+    /**
+     * Lookup function for the SD-JWT VC Type Metadata of a `vct`.
+     */
     val lookupTypeMetadata: LookupTypeMetadata
 
+    /**
+     * Maximum number of parent lookups allowed by this resolver.
+     */
+    val limit: UInt?
```
babisRoutis: Remove this. Breaks the public API without a good reason.

Consider the following:

```kotlin
interface ResolveTypeMetadata {

    /**
     * Lookup function for the SD-JWT VC Type Metadata of a `vct`.
     */
    val lookupTypeMetadata: LookupTypeMetadata

    /**
     * Resolves the [ResolvedTypeMetadata] for [vct].
     */
    suspend operator fun invoke(
        vct: Vct,
        expectedIntegrity: DocumentIntegrity?,
    ): Result<ResolvedTypeMetadata>

    companion object {
        operator fun invoke(
            lookupTypeMetadata: LookupTypeMetadata,
            limit: UInt? = null,
        ): ResolveTypeMetadata {
            if (null != limit) {
                require(limit > 0u) { "limit cannot be 0" }
            }

            return object : ResolveTypeMetadata {
                override val lookupTypeMetadata: LookupTypeMetadata = lookupTypeMetadata
                override suspend fun invoke(
                    vct: Vct,
                    expectedIntegrity: DocumentIntegrity?,
                ): Result<ResolvedTypeMetadata> =
                    runCatchingCancellable {
                        tailrec suspend fun resolve(
                            vct: Vct,
                            expectedIntegrity: DocumentIntegrity?,
                            accumulator: ResolvedTypeMetadata,
                            resolved: Set<Vct>,
                        ): ResolvedTypeMetadata {
                            require(vct !in resolved) { "cyclical reference detected, vct $vct has been previously resolved" }
                            limit?.let {
                                require(resolved.size <= it.toInt()) { "maximum number of parent lookups exceeded, limit: $it" }
                            }
                            val current = lookupTypeMetadata(vct, expectedIntegrity)
                                .getOrThrow() ?: error("unable to lookup Type Metadata for $vct")
                            val updatedAccumulator = accumulator + current
                            val parent = current.extends?.let { Vct(it) }
                            val parentIntegrity = current.extendsIntegrity
                            return if (null != parent) {
                                resolve(parent, parentIntegrity, updatedAccumulator, resolved + vct)
                            } else {
                                updatedAccumulator
                            }
                        }

                        resolve(vct, expectedIntegrity, ResolvedTypeMetadata.empty(vct), emptySet())
                    }
            }
        }
    }
}
```

It just removes the default implementation of the method

babisRoutis (changes requested):

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
dzarras: Done. Extracted the `tailrec` function to avoid duplication, and updated based on your suggestion.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
```diff
@@ -100,33 +98,71 @@ interface ResolveTypeMetadata {
         expectedIntegrity: DocumentIntegrity?,
     ): Result<ResolvedTypeMetadata> =
         runCatchingCancellable {
```
babisRoutis, line 100: Why implement this?

It allows the resolution without a limit, keeping the problem present

If for whatever reason this needs to be implement here, then add a "reasonable" limit (4)

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
```diff
@@ -100,33 +98,71 @@ interface ResolveTypeMetadata {
...
+        }
+    }
+}
+
+private tailrec suspend fun LookupTypeMetadata.resolve(
+    vct: Vct,
+    expectedIntegrity: DocumentIntegrity?,
+    accumulator: ResolvedTypeMetadata,
+    resolved: Set<Vct>,
+    limit: UInt?,
```
babisRoutis: Make it, non-nullable

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
```diff
@@ -100,33 +98,71 @@ interface ResolveTypeMetadata {
...
+
+        /**
+         * Creates a new [ResolveTypeMetadata] instance.
+         *
+         * @param lookupTypeMetadata lookup function for the SD-JWT VC Type Metadata of a `vct`
+         * @param limit maximum number of parent lookups allowed by this resolver; when `null`, no limit is enforced
+         */
+        operator fun invoke(
+            lookupTypeMetadata: LookupTypeMetadata,
+            limit: UInt? = null,
```
babisRoutis: Make it no null with a default value of 4

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
dzarras: If we remove the default implementation, then we break existing clients that opted to implement the interface rather than use the factory method of the companion.

I'll add a default limit as suggested.

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
dzarras: Done

src/main/kotlin/eu/europa/ec/eudi/sdjwt/vc/ResolveTypeMetadata.kt
dzarras: Done

babisRoutis (approved):

PR #509, opened 2026-08-31
Rework `SdJwtDefinition.fromSdJwtVcMetadata()` using `DeepRecursiveFunction`

### LINKED ISSUE #493: `SdJwtDefinition.Companion.fromSdJwtVcMetadata()` can result in `StackOverflowError`
SD-JWT VC Type Metadata documents are not bound.

Thus parsing a sufficiently large document into an `SdJwtDefinition` can result in a `StackOverflowError` being thrown due to the recursive nature of the function.

We must update the implementation to make use of `DeepRecursiveFunction` and avoid such issues.

dzarras:
Closes #493


src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/FromSdJwtVcMetadata.kt
```diff
@@ -37,102 +37,104 @@ fun SdJwtDefinition.Companion.fromSdJwtVcMetadata(
...
+    val childClaimsMetadatas: List<ClaimMetadata>,
+    val allClaimsGroupedByParentPath: Map<ClaimPath?, List<ClaimMetadata>>,
+    val selectivelyDiscloseWhenAllowed: Boolean,
+)
+
+private typealias SdJwtObjectDefinitionContent = Map<String, SdJwtElementDefinition>
+
+private val processObjectDefinition: DeepRecursiveFunction<SdJwtObjectDefinitionContext, SdJwtObjectDefinitionContent> =
+    DeepRecursiveFunction { (childClaimsMetadatas, allClaimsGroupedByParentPath, selectivelyDiscloseWhenAllowed) ->
+        suspend fun DeepRecursiveScope<*, *>.metaOf(
```
babisRoutis: Minor

This nested function can be inlined 

@dzarras Keep the version that you prefer

babisRoutis (dismissed):

src/main/kotlin/eu/europa/ec/eudi/sdjwt/dsl/def/FromSdJwtVcMetadata.kt
dzarras: Removed the nested functions.

